PR #296, opened 2024-09-11
Add pipeline for node auto provisioning

anson627:


modules/python/autoscale/autoscale.py
```diff
@@ -23,36 +24,17 @@ def check_count(desired_count, command, resource_type, timeout, start_time):
 
     return False, -1
 
-def calculate_request_resource(node_name, node_count, pod_count, input_file, output_file):
-    allocatable_cpu = run_command(f"kubectl get node {node_name} -o jsonpath='{{.status.allocatable.cpu}}'")
```
alyssa1303, line 27: I think this assumption is okay for now, but in the future, it's best if we can get a node name (a node with same sku in default nodepool) to get its actual allocatable cpu & memory. In AKS, different cluster versions have different values for kube-reserved, and the same goes for AWS since we don't know the actual reserved value.

anson627: agree, the problem is currently we use the allocatable resources from system node pool, which might be different from the user node pool, also it's hard to get node name, in future we might want to warmup the test a bit to get one user node up, and then use the allocatable to calculate test pod resources correctly

PR #297, opened 2024-09-11
Add script to calculate scale up latency

alyssa1303:


modules/python/autoscale/autoscale.py
```diff
@@ -0,0 +1,120 @@
...
+    with open(input_file, 'r') as file:
+        content = file.read()
+
+    content = content.replace("##CPUperJob##", f"{cpu_request}{cpu_unit}")
+    content = content.replace("##Replicas##", str(pod_count))
+
+    with open(output_file, 'w') as file:
+        file.write(content)
+
+def run_deployment(yaml_file, pod_count, result_file):
```
anson627, line 48: run_jobs?

modules/python/autoscale/deployment_template.yaml
```diff
@@ -0,0 +1,43 @@
...
+            requests:
+              cpu: ##CPUperJob##
+            limits:
+              cpu: ##CPUperJob##
+          command:
+            - "/bin/bash"
+            - "-c"
+            - runtime="1 minute"; endtime=$(date -ud "$runtime" +%s); while [[ $(date -u +%s) -le $endtime ]]; do echo $(date); sleep 6000; done
+      restartPolicy: Always
+      tolerations:
```
anson627, line 30: the tolerations and affinity only work on azure, we probably want to expose variables for aws to make this template generic

modules/python/autoscale/autoscale.py
alyssa1303: I switch to use deployment instead of batch jobs but sure I can rename the function to run_jobs

modules/python/autoscale/autoscale.py
```diff
@@ -0,0 +1,120 @@
...
+    with open(output_file, 'w') as file:
+        file.write(content)
+
+def run_jobs(yaml_file, pod_count, result_file):
+    start_time = time.time()
+    print(f"Start time: {start_time}")
+    run_command(f"kubectl apply -f {yaml_file}")
+
+    # Wait for all pods to be running
+    while True:
```
anson627, line 54: can we also check desired number of nodes are ready as well? can add wait_for_nodes_seconds

modules/python/autoscale/autoscale.py
```diff
@@ -0,0 +1,120 @@
...
+        if pods_num == pod_count:
+            break
+        print(f"Waiting for all pods to be running, current pods count: {pods_num}")
+        time.sleep(5)
+
+    end_time = time.time()
+    print(f"End time: {end_time}")
+
+    run_time = end_time - start_time
+    print(f"Total run time in seconds: {run_time}")
```
anson627, line 65: wait_for_pods_seconds?

modules/python/autoscale/autoscale.py
```diff
@@ -0,0 +1,120 @@
...
+
+def run_jobs(yaml_file, pod_count, result_file):
+    start_time = time.time()
+    print(f"Start time: {start_time}")
+    run_command(f"kubectl apply -f {yaml_file}")
+
+    # Wait for all pods to be running
+    while True:
+        pods_num = int(run_command("kubectl get pods | grep scale-deployment | grep -c Running"))
+        if pods_num == pod_count:
```
anson627, line 56: need a timeout say 30 minutes, and mark test as failure

anson627 (approved):

PR #304, opened 2024-09-17
Implement clusterloader2 to run autoscale test

alyssa1303:
Implement ClusterLoader2 to run autoscale test


modules/python/clusterloader2/autoscale/autoscale.py
```diff
@@ -0,0 +1,107 @@
...
+    calculate_request_resource(cpu_per_node, node_count, pod_count, cl2_override_file)
+    result = run_cl2_command(kubeconfig, cl2_config_dir, cl2_report_dir, provider, overrides=True)
+    print(result)
+
+def collect_clusterloader2(cpu_per_node, node_count, pod_count, autoscale_type, cl2_report_dir, cloud_info, run_id, run_url, result_file):
+    raw_data = parse_xml_to_json(os.path.join(cl2_report_dir, "junit.xml"))
+    json_data = json.loads(raw_data)
+    testsuites = json_data["testsuites"]
+    wait_for_nodes_seconds = -1
+    wait_for_pod_seconds = -1
```
anson627, line 38: wait_for_pods_seconds? and I probably need update the schema

modules/python/clusterloader2/utils.py
```diff
@@ -0,0 +1,89 @@
+import json
+import subprocess
+import os
+import docker
+
+from xml.dom import minidom
+from docker_client import DockerClient
+
+IMAGE="telescope.azurecr.io/perf-eval/clusterloader2:20240917.1"
```
anson627, line 9: can we make this configurable from pipeline?

modules/python/clusterloader2/autoscale/autoscale.py
```diff
@@ -0,0 +1,107 @@
...
+    cpu_request = (cpu_per_node * 1000 * 0.9) * node_count // pod_count
+
+    print(f"Total number of nodes: {node_count}, total number of pods: {pod_count}")
+    print(f"CPU request for each pod: {cpu_request}m")
+
+    with open(override_file, 'r') as file:
+        content = file.read()
+
+    # assuming then number of surge nodes is no more than 10
+    content = content.replace("##CPUperJob##", f"{cpu_request}m")
```
anson627, line 20: can we generate the overrides yaml file in python instead of replacing string

steps/engine/clusterloader2/autoscale/execute.yml
```diff
@@ -0,0 +1,27 @@
...
+  type: object
+  default: {}
+- name: region
+  type: string
+
+steps:
+- script: |
+    set -eo pipefail
+
+    PYTHONPATH=$PYTHONPATH:$(pwd) python3 $PYTHON_SCRIPT_FILE execute $CPU_PER_NODE $NODE_COUNT $POD_COUNT \
```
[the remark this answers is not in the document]
alyssa1303: We do because it's needed to call dependency from 1 file to another inside clusterloader2 modules. What do you mean by setting it for all pipelines?

modules/python/clusterloader2/autoscale/autoscale.py
alyssa1303: So you mean to get rid of the overrides.yaml file?

modules/python/clusterloader2/utils.py
alyssa1303: done

anson627 (dismissed):

alyssa1303:
> what is the difference btw failures and errors, should we include as part of results for debugging purpose?

If you're talking about docker error and test suite failure, they're the same. When the test suit has failure, docker process will exit with error code 1, so I have to handle it gracefully to not fail the pipeline. Test suite failure means either node is not ready or pod is not running in our autoscale case. I already logged everything to the terminal and include raw data in the result so I don't think there's more to logs.

PR #309, opened 2024-09-18
Add env variable RUN_URL, TEST_RESULTS_DIR and TEST_RESULTS_FILE

anson627:


sumanthreddy29 (dismissed):

alyssa1303 (commented):
Can you make a change to update the order of RUN_ID to this instead `run_id=$(Build.BuildId)-$(System.JobId)`

anson627:
> Can you make a change to update the order of RUN_ID to this instead `run_id=$(Build.BuildId)-$(System.JobId)`

Done

steps/cloud/azure/upload-storage-account.yml
```diff
@@ -28,7 +28,7 @@ steps:
     fi
   displayName: "Upload Result"
   env:
-    SOURCE_FILE_NAME: ${{ parameters.source_file_name }}
+    SOURCE_FILE_NAME: $(TEST_RESULTS_FILE)
```
sumanthreddy29, line 31:
```suggestion
    SOURCE_FILE_NAME: ${{ parameters.source_file_name }}
```
can we revert this change, so that we allow user to override the file name ?

sumanthreddy29: If we don't want user to override, we might have to remove parameter and use the env variable `TEST_RESULTS_FILE` in all the places

anson627: reverted, nice catch!

steps/engine/kperf/execute.yml
```diff
@@ -79,7 +77,6 @@ steps:
     BENCHMARK_SUBCMD: ${{ parameters.engine_input.benchmark_subcmd }}
     BENCHMARK_SUBCMD_ARGS: ${{ parameters.engine_input.benchmark_subcmd_args }}
     EXTRA_BENCHMARK_SUBCMD_ARGS: ${{ parameters.extra_benchmark_subcmd_args }}
-    RESULT_DIR: $(System.DefaultWorkingDirectory)/$(RUN_ID)
+    RESULT_DIR: $(TEST_RESULTS_DIR)
```
alyssa1303, line 80: Maybe you can use TEST_RESULTS_DIR directly in the script like the collect step and get rid of this env var?

sumanthreddy29 (approved):

PR #318, opened 2024-10-02
Refactor Terraform run template

sumanthreddy29:


steps/terraform/run-command.yml
```diff
@@ -43,38 +43,6 @@ steps:
             az resource delete --ids $ids --verbose
             rm -r terraform.tfstate.d/$region
           fi
-          if [[ ${{ parameters.command }} == "destroy" && "$CLOUD" == "aws" ]]; then
```
alyssa1303, line 46: I think we should just remove it completely. This is included as part of k8s storage test but they already make a template to run this clean up.

sumanthreddy29: Okay. I will remove it completely

alyssa1303 (approved):

PR #336, opened 2024-10-14
Add provisioner to cleanup any available attached ENIs during subnet destroy

rafael-mendes-pereira:
This is a workaround for the known VPC CNI addon's "leaked ENIs" issue: See https://github.com/aws/amazon-vpc-cni-k8s/issues/608

### DIFF
--- a/modules/terraform/aws/virtual-network/main.tf
+++ b/modules/terraform/aws/virtual-network/main.tf
@@ -32,6 +32,32 @@ resource "aws_subnet" "subnets" {
   tags = merge(local.tags, {
     "Name" = each.value.name
   })
+
+  provisioner "local-exec" {
+    /*
+      This command cleans up any not in use ENIs attached to this subnet, which were created outside the scope of the Terraform modules.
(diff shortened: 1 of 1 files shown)


modules/terraform/aws/virtual-network/main.tf
```diff
@@ -32,6 +32,32 @@ resource "aws_subnet" "subnets" {
...
   })
+
+  provisioner "local-exec" {
+    /*
+      This command cleans up any not in use ENIs attached to this subnet, which were created outside the scope of the Terraform modules.
+      This is a workaround for the known VPC CNI addon's "leaked ENIs" issue: See https://github.com/aws/amazon-vpc-cni-k8s/issues/608
+    */
+    when    = destroy
+    command = <<-EOT
+      echo "Detaching Subnet: ${self.id} Network Interfaces"
```
sumanthreddy29, line 43: This only helps with leaked eni's. We also need to delete eks security that creates dependency on vpc.

sumanthreddy29 (approved):

modules/terraform/aws/virtual-network/main.tf
sumanthreddy29: Rafael will add this change as part of another PR.

PR #339, opened 2024-10-16
Use aws provider default_tags

rafael-mendes-pereira:
Introduce the [aws provider default_tags](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/resource-tagging#propagating-tags-to-all-resources) to propagate the required tags to all supported resources.  

This feature is important to enforce tags, such as, `owner` and `deletion_due_time` , which are used for the cost tracking and garbage collection.

Note: this feature has some limitations: 
1. No support to [aws_autoscaling_group](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_group)
2. No support to computed values (e.g. `timestamp()`). See Terraform issue [#19583](https://github.com/hashicorp/terraform-provider-aws/issues/19583#issuecomment-1561337902). 
To overtake this issue, this change gets the current time from the terraform input variables and stop using  `timestamp()` (note that Terraform [does not recommend](https://developer.hashicorp.com/terraform/language/functions/timestamp) to use `timestamp()` in resources attributes)


anson627 (dismissed):

anson627:
need fix failed terraform validation

rafael-mendes-pereira:
> need fix failed terraform validation

Done. Thanks

.github/workflows/terraform-validation.yml
```diff
@@ -144,7 +144,7 @@ jobs:
       - name: Create JSON Input
         run: |
           INPUT_JSON_OBJECT=$(jq . ${{ env.TERRAFORM_TEST_INPUT_FILE }})
-          INPUT_JSON_STRING=$(echo $INPUT_JSON_OBJECT | jq -c .)
+          INPUT_JSON_STRING=$(echo $INPUT_JSON_OBJECT | jq -c '. + {current_time: (now |  todateiso8601)}')
           echo "INPUT_JSON=$INPUT_JSON_STRING" >> "$GITHUB_ENV"
```
alyssa1303, line 148: Maybe set this like RUN_ID?

modules/terraform/aws/README.md
```diff
@@ -53,16 +53,17 @@ INPUT_JSON=$(jq -n \
       '{
       run_id: $run_id,
       region: $region
-      }' | jq 'with_entries(select(.value != null and .value != ""))')
```
alyssa1303, line 56: Same here. Set in like RUN_ID and use as arg instead of extending?

modules/terraform/aws/README.md
```diff
@@ -53,16 +53,17 @@ INPUT_JSON=$(jq -n \
...
+      }'  | jq '. + {current_time: (now |  todateiso8601)}' \
+          | jq 'with_entries(select(.value != null and .value != ""))')
 ```
 **Note**: The `jq` command will remove any null or empty values from the JSON object. So any variable surrounded by double quotes means it is optional and can be removed if not needed.
 
 ### Provision resources using Terraform:
 ```bash
 pushd $TERRAFORM_MODULES_DIR
 terraform init
-terraform plan -var json_input=$(echo $INPUT_JSON | jq -c .) -var-file $TERRAFORM_INPUT_FILE
```
alyssa1303, line 64: While are they needed here if INPUT_JSON already has current_time?

modules/terraform/aws/main.tf
```diff
@@ -1,15 +1,21 @@
...
 
-  tags = {
-    "owner"             = var.owner
+  non_computed_tags = {
+    # Note: Define only non computed values (i.e. values that do not change for each resource). This is required due to a limitation at "aws" provider default_tags.
+    "owner"             = var.owner # note: MUST NOT REMOVE (it's used for resources accountability and cost tracking)
     "scenario"          = "${var.scenario_type}-${var.scenario_name}"
-    "creation_time"     = timestamp()
-    "deletion_due_time" = timeadd(timestamp(), var.deletion_delay)
+    "creation_time"     = local.current_time                              # note: should not use timestamp() since it is a computed value
```
alyssa1303, line 10: Can it be part of json_input like run_id?

alyssa1303 (commented):
We should do a test run with internal pipelines to make sure no breaking change

steps/terraform/run-command.yml
```diff
@@ -34,7 +34,11 @@ steps:
         terraform_input_variables=$(echo $TERRAFORM_REGIONAL_CONFIG | jq -r --arg region "$region" '.[$region].TERRAFORM_INPUT_VARIABLES')
 
         set +e
-        terraform ${{ parameters.command }} --auto-approve ${{ parameters.arguments }} -var-file $terraform_input_file -var json_input=$terraform_input_variables
+        if [[ "$CLOUD" == "aws" ]]; then
```
alyssa1303, line 37: I think we're trying to avoid doing this if possible since this is a shared step. What I meant by not appending to json_input is to first define it, then use it as an argument like this
```
CURRENT_TIME=$(date -uIseconds | sed 's/+00:00/Z/')
INPUT_JSON=$(jq -n \
      --arg run_id $RUN_ID \
      --arg current_time $CURRENT_TIME \
      --arg region $REGION \
      '{
      run_id: $run_id,
      current_time: $current_time,
      region: $region
      }' | jq 'with_entries(select(.value != null and .value != ""))')
```

So still the old way but do it like RUN_ID instead of appending after jq. Sorry if my explanation wasn't clear earlier.

alyssa1303 (approved):

modules/terraform/aws/karpenter/main.tf
```diff
@@ -64,13 +64,6 @@ locals {
   eks_managed_node_group  = var.karpenter_config.eks_managed_node_group
   karpenter_chart_version = var.karpenter_config.karpenter_chart_version
 
-  tags = {
```
sumanthreddy29, line 67: these tags are for open source karpenter module. We will keep the same way they are right now as we are not using it currently.

modules/terraform/aws/karpenter/variables.tf
```diff
@@ -23,27 +23,3 @@ variable "json_input" {
     region = string
   })
 }
-
-variable "owner" {
```
sumanthreddy29, line 27: These are required for karpenter open source module as this module becomes the main module when we use open source karpenter.

modules/terraform/aws/virtual-network/main.tf
```diff
@@ -9,7 +9,7 @@ locals {
   nat_gateways_map             = var.network_config.nat_gateways == null ? {} : { for ng in var.network_config.nat_gateways : ng.name => ng }
 
   security_group_name = var.network_config.security_group_name
-  tags                = merge(var.tags, { "role" = var.network_config.role })
```
sumanthreddy29, line 12: In networking module, we still have lot tag references

sumanthreddy29 (changes requested):

steps/terraform/run-command.yml
rafael-mendes-pereira: As we talked in Teams, we need to make sure that the current time is updated when the resources are provisioned. Having it in the input json would require to update the json every time someone runs 'terraform apply'. To prevent that and make the code easier to maintain, I added the `current_time` as another input variable for terraform commands.

PR #346, opened 2024-10-21
VPC-CNI: enable prefix delegation AND remove ENIs cleanup 

rafael-mendes-pereira:
Enabling IPv4 prefix delegation to increase number of PODs per ENIs.

We've seen that it prevents the ENIs leak. Therefore, we don't need the ENI cleanup scripts anymore.


steps/engine/clusterloader2/autoscale/cleanup-aws.yml
```diff
@@ -1,41 +1,6 @@
 steps:
 - script: |
-
     kubectl delete --all nodeclaim --ignore-not-found
     kubectl delete --all nodepool --ignore-not-found
```
alyssa1303, line 4: Since we remove this, we can combine the cleanup-aws.yaml and cleanup-azure.yml into a single file like before.

PR #361, opened 2024-10-24
Update config to support 1000 nodes for cilium scale test

alyssa1303:
Summary:
- Update Azure network and aks modules to support customize values to create cluster with 1000 nodes
- Update AWS module to over configuration values of addons to work around scaling to 50k pods issue
- Update test to run on 1000 nodes instead of 100 nodes


pipelines/perf-eval/slo-nodes100-pods5000.yml
```diff
@@ -1,41 +1,41 @@
...
           regions:
-            - us-east-2
+            - us-east-1
           engine: clusterloader2
           engine_input:
-            image: "ghcr.io/azure/clusterloader2:v20241016"
+            image: "ghcr.io/azure/clusterloader2:v20241022"
           topology: slo
           matrix:
             n100-p5000:
```
sumanthreddy29, line 29:
```suggestion
            n1000-p50000:
```

alyssa1303: actually I think I'll update to aws_vpc_cni to match naming on azure side

anson627 (approved):

PR #367, opened 2024-10-29
Add Extended Zone API server benchmark test scenario

viananth:
This PR adds a new test scenario for Extended Zone "losangeles" to test API server benchmark for 10 Virtual Nodes and 100 pods.


pipelines/perf-eval/apiserver-benchmark-virtualnodes10-pods100-EZ.yml
```diff
@@ -0,0 +1,40 @@
...
+            exempt:
+              flowcontrol: "exempt:5"
+              extra_benchmark_subcmd_args: ""
+          engine_input:
+            runner_image: telescope.azurecr.io/oss/kperf:v0.1.3
+            benchmark_subcmd: node10_job1_pod100
+            benchmark_subcmd_args: "--total 1000"
+          max_parallel: 1
+          timeout_in_minutes: 360
+          credential_type: service_connection
```
sumanthreddy29, line 40: can you add  ` ssh_key_enabled: false` since we won;t be creating SSH keys for this test

sumanthreddy29 (approved):

PR #370, opened 2024-10-30
Add secondary cidr block to vpc to increase subnets

alyssa1303:
Summary
- Add secondary cidr block to vpc to increase subnet number and range
- Change back the order of addons


anson627 (dismissed):

sumanthreddy29:
Resolve conflicts

modules/terraform/aws/variables.tf
```diff
@@ -38,9 +38,10 @@ variable "deletion_delay" {
 variable "network_config_list" {
   description = "Configuration for creating the server network."
   type = list(object({
-    role           = string
-    vpc_name       = string
-    vpc_cidr_block = string
+    role                       = string
+    vpc_name                   = string
+    vpc_cidr_block             = string
+    secondary_ipv4_cidr_blocks = optional(list(string))
```
sumanthreddy29, line 44: can we add default value as empty list or null? Also, can we test this new variable from internal repo as the internal pipelines use this module?

alyssa1303: Will test internally. I already check for null value in the mapping thou, and other variables don't have default values either

sumanthreddy29 (approved):

modules/terraform/aws/virtual-network/main.tf
```diff
@@ -32,6 +40,8 @@ resource "aws_subnet" "subnets" {
   tags = merge(local.tags, {
     "Name" = each.value.name
   })
+
+  depends_on = [aws_vpc_ipv4_cidr_block_association.secondary_ipv4_cidr_block]
```
rafael-mendes-pereira, line 44: Terraform [recommends](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/subnet#subnets-in-secondary-vpc-cidr-blocks) to refer to the VPC resource id in the specific subnet of the additional CIDR block. If we have the subnet as another resource, we won't need to add it in the VPC's subnet list config. This would be:

```
resource "aws_subnet" "in_secondary_cidr" {
  for_each = local.secondary_ipv4_cidr_block_map

  vpc_id     = aws_vpc_ipv4_cidr_block_association.secondary_ipv4_cidr_block[each.key].vpc_id
  cidr_block = each.value
}
```

alyssa1303: We can do another refactor task if a new resource is really necessary. What Terraform recommends can be achieved the same way as this approach since both ensuring secondary cidr blocks are created before subnets are created

rafael-mendes-pereira (approved):

rafael-mendes-pereira (approved):

PR #371, opened 2024-10-30
Cluster Auto Scaler Test case

sumanthreddy29:
- Add Support for Cluster Auto scaler for AWS and Azure.
- Update `autoscaler.py` to parse node selector and node label selector from the command line arguments.
- Added new test case for cluster auto scaler with 10 Nodes and 100 Pods similar to Karpenter.
- For Cluster auto scaler need OIDC provider, so moved OIDC provider to EKS module instead of EKS addon module.


modules/terraform/aws/eks/cluster-autoscaler/autoscaler.yml
```diff
@@ -0,0 +1,196 @@
...
+      containers:
+        - image: registry.k8s.io/autoscaling/cluster-autoscaler:${IMAGE_TAG}
+          name: cluster-autoscaler
+          resources:
+            limits:
+              cpu: 100m
+              memory: 300Mi
+            requests:
+              cpu: 50m
+              memory: 200Mi
```
alyssa1303, line 158: What are the replica and resources values based on? The [doc](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml) seems to suggest different values

scenarios/perf-eval/cas-c4n10p100/terraform-inputs/aws.tfvars
```diff
@@ -0,0 +1,104 @@
...
+  eks_managed_node_groups = [
+    {
+      name           = "cas-c4n10p100-ng"
+      ami_type       = "AL2_x86_64"
+      instance_types = ["m4.large"]
+      min_size       = 5
+      max_size       = 5
+      desired_size   = 5
+      capacity_type  = "ON_DEMAND"
+      labels         = { "autoscaler" = "owned" }
```
alyssa1303, line 84: Why is this label needed? I don't think we enable any autoscale for karpenter case

scenarios/perf-eval/cas-c4n10p100/terraform-inputs/aws.tfvars
```diff
@@ -0,0 +1,104 @@
...
+
+eks_config_list = [{
+  role                      = "cas"
+  eks_name                  = "cas-c4n10p100"
+  enable_cluster_autoscaler = true
+  vpc_name                  = "cas-vpc"
+  policy_arns               = ["AmazonEKSClusterPolicy", "AmazonEKSVPCResourceController", "AmazonEKSWorkerNodePolicy", "AmazonEKS_CNI_Policy", "AmazonEC2ContainerRegistryReadOnly", "AmazonSSMManagedInstanceCore"]
+  eks_managed_node_groups = [
+    {
+      name           = "cas-c4n10p100-ng"
```
alyssa1303, line 77: Maybe rename to "default" for consistency with Azure

modules/terraform/azure/variables.tf
```diff
@@ -165,6 +165,25 @@ variable "aks_config_list" {
     role_assignment_list = optional(list(string), [])
     kubernetes_version   = optional(string, null)
     edge_zone            = optional(string, null)
+    auto_scaler_profile = optional(object({
```
alyssa1303, line 168: Need to test Cilium pipeline to ensure no breaking change on Azure side

scenarios/perf-eval/cas-c4n10p100/terraform-inputs/aws.tfvars
sumanthreddy29: Based on this label we install the auto scaler on that node.

modules/terraform/aws/eks/cluster-autoscaler/autoscaler.yml
sumanthreddy29: I updated the resource values so that we can spin up 10 nodes for 100 pods or else only 5 nodes will spin up when we create the test deployment

scenarios/perf-eval/cas-c4n10p100/terraform-inputs/aws.tfvars
sumanthreddy29: Updated

scenarios/perf-eval/cas-c4n10p100/terraform-inputs/aws.tfvars
sumanthreddy29: Removed Labels for aws

modules/terraform/aws/eks/cluster-autoscaler/autoscaler.yml
alyssa1303: Then might as well test this on creating 1000 nodes if we plan to replace karpenter with autoscale

sumanthreddy29: Anson suggested we monitor the results first and then later on we can replace the karpenter with Autoscale.

sumanthreddy29: I already tested this on slo test case before with 100 nodes and It worked.
https://dev.azure.com/akstelescope/telescope/_build/results?buildId=2906&view=logs&j=e5292d76-0a0c-5aec-1dff-3e42152c38be&t=1e699bf3-59fa-51ba-8f55-d44a80bf3655

modules/terraform/aws/eks/cluster-autoscaler/autoscaler.yml
```diff
@@ -0,0 +1,195 @@
...
+          command:
```
alyssa1303: I think we should use value that can support 1k nodes so user don't have to modify it based on nodes. Also, it doesn't seem like we currently allow the customization of the value if there are different test cases using auto-scaler with different node counts

pipelines/perf-eval/cilium-cluster-churn-nodes1000-pods50k.yml
```diff
@@ -61,8 +61,8 @@ stages:
               cilium_enabled: False
             azure_cilium:
               cpu_per_node: 4
-              node_count: 1000
```
alyssa1303, line 64: revert change

sumanthreddy29: Reverted

alyssa1303 (approved):

PR #379, opened 2024-11-01
Move current_time to input json and rename it to creation_time

rafael-mendes-pereira:
Instead of having `current_time` as another input variable for terraform, move it to json input, therefore, it won't affect how we run the terraform command for each cloud type. It also renames it to `creation_time` as this is the actual meaning of this variable.


modules/terraform/aws/main.tf
```diff
@@ -1,13 +1,14 @@
 locals {
-  region = lookup(var.json_input, "region", "us-east-1")
-  run_id = lookup(var.json_input, "run_id", "123456")
+  region        = lookup(var.json_input, "region", "us-east-1")
+  run_id        = lookup(var.json_input, "run_id", "123456")
+  creation_time = var.json_input["creation_time"]
```
sumanthreddy29, line 4:
```suggestion
  creation_time = lookup(var.json_input, "creation_time", null )
```
Can we use the lookup syntax here also. With default value as current time incase if the user doesn't pass the time

modules/terraform/aws/README.md
```diff
@@ -21,13 +21,15 @@ SCENARIO_NAME=nap-c4n10p100
...
 REGION="us-east-2"
+CREATION_TIME=$(date -uIseconds |  sed 's/+00:00/Z/')
 TERRAFORM_MODULES_DIR=$(pwd)/modules/terraform/$CLOUD
 TERRAFORM_INPUT_FILE=$(pwd)/scenarios/$SCENARIO_TYPE/$SCENARIO_NAME/terraform-inputs/${CLOUD}.tfvars
 ```
 
 **Note**:
 
 * `RUN_ID` should be a unique identifier since it is used to identify the resources based on tags as AWS has no concept of a resource group.
+* `CRETION_TIME` is used to define the tags `creation_time` and `deletion_due_time`, which can be used to track provisioned resources in the AWS account.  It should not be more or less than one hour from the time the commands terraform plan/apply are run.
```
sumanthreddy29, line 32: `CREATION_TIME` typo

sumanthreddy29:
Can you update the aws input schema json also?

modules/terraform/aws/aws_input_schema.json
```diff
@@ -6,11 +6,15 @@
...
+    },
+    "creation_time": {
+      "type": "string"
     }
   },
   "required": [
     "run_id",
-    "region"
+    "region",
+    "creation_time"
```
sumanthreddy29, line 17: Remove this as required as we add this value in the workflow. Also making it required will fail all the aws test as we need to update all the tests json's with creation_time.

sumanthreddy29 (approved):

PR #380, opened 2024-11-04
Add Telescope Metadata to the results

sumanthreddy29:
1. Collect Telescope Metadata and upload.
2. Update final results schema to have metadata and results.


pipelines/perf-eval/node-auto-provisioning-benchmark-nodes10-pods100.yml
```diff
@@ -10,7 +10,7 @@ schedules:
 variables:
   SCENARIO_TYPE: perf-eval
   SCENARIO_NAME: nap-c4n10p100
-  SCENARIO_VERSION: main
```
sumanthreddy29, line 13: Revert test changes

steps/cloud/azure/upload-storage-account.yml
```diff
@@ -30,6 +30,9 @@ steps:
   env:
     SOURCE_FILE_NAME: ${{ parameters.source_file_name }}
     DESTINATION_FILE_NAME: ${{ parameters.destination_file_name }}
-    SUBFOLDER: ${{ parameters.subfolder }}
     CONTAINER_NAME: ${{ parameters.container_name}}
     AZURE_STORAGE_ACCOUNT_NAME: $(AZURE_STORAGE_ACCOUNT_NAME)
+    ${{ if ne(variables['Build.SourceBranchName'], 'main') }}:
+      SUBFOLDER: $(SCENARIO_NAME)/$(Build.SourceBranchName)
+    ${{ else }}:
+      SUBFOLDER: ${{ parameters.subfolder }}
```
alyssa1303, line 38: I don't think it's a good idea. Doesn't this mean the SCENARIO_VERSION is no longer usable? It'll confuse users and they can't find where to look for blobs to check results correctly. For example, when someone sets scenario_version to a tag, it just got push to somewhere else? This also means that we no longer allow the flexibility of using something else other than 'main' as scenario_version for scheduled run? What if users want to change schema, do they always have to delete and recreate the table?

steps/collect-telescope-metadata.yml
```diff
@@ -0,0 +1,81 @@
...
+    ENGINE: ${{ parameters.engine }}
+    TOPOLOGY: ${{ parameters.topology }}
+    ENGINE_IMAGE: ${{ parameters.image }}
+    CLOUD: ${{ parameters.cloud }}
+    REGIONS: ${{ convertToJson(parameters.regions) }}
+
+- script: |
+    # Read the test results file and add the telescope metadata to it for each json record in the file
+    set -eux
+    jq --argfile telescope_metadata $(TELESCOPE_METADATA_FILE) \
```
alyssa1303, line 68: I recommend testing it with test that produce a big jq result file like compute or iperf3 to see if jq can handle it or not

steps/setup-tests.yml
```diff
@@ -48,6 +52,26 @@ steps:
     echo "##vso[task.setvariable variable=TEST_RESULTS_FILE]$test_results_file"
   displayName: 'Set Test Results Directory'
 
+- script: |
+    telescope_metadata=$(jq -n \
```
alyssa1303, line 56: Why do we have to set it here?

steps/collect-telescope-metadata.yml
```diff
@@ -0,0 +1,81 @@
...
+      -c '. + {telescope_metadata: $telescope_metadata}' $(TEST_RESULTS_FILE) > temp-$RUN_ID.json \
+       && mv temp-$RUN_ID.json $(TEST_RESULTS_FILE)
+  displayName: "Add Telescope Metadata to Test Results"
+  condition: always()
+
+- ${{ if eq(variables['Build.SourceBranchName'], 'add-telescope-metadata') }}:
```
alyssa1303, line 74: What is this step for?

sumanthreddy29: We don't want to collect data that is not from main branch. For testing I added my branch here.

steps/setup-tests.yml
sumanthreddy29: Removed

steps/cloud/azure/upload-storage-account.yml
sumanthreddy29: 1. `SCENARIO_VERSION` is used when we run a manual/scheduled trigger against main branch which is part of else statement. 
2. If someone sets `scenario_version`as a tag, we push the data to tag folder itself if the run is triggered from main branch. 
3. Let's say when user creates a test branch and uses that test branch name as `scenario_version` with scheduled run we push the data to the branch which triggered the pipeline. Only restriction is they have to set the scenario version and branch name same here which allows the flexibility of using any value for `scenario_version`

steps/collect-telescope-metadata.yml
alyssa1303: Why don't we want it? Doesn't that mean results we see in testing is different from testing we see in main? I think we should make it consistent so people know for sure what to expect in the data

steps/cloud/azure/upload-storage-account.yml
sumanthreddy29: reverted this change

steps/collect-telescope-metadata.yml
sumanthreddy29: this step only uploads metadata for the run which we use for building the dashboard. I thought it's better not to collect.

steps/collect-telescope-metadata.yml
sumanthreddy29: Tested with compute and iperf test cases. Updated the internal post with the links

steps/collect-telescope-metadata.yml
```diff
@@ -0,0 +1,82 @@
...
+steps:
+- script: |
+    telescope_metadata=$(jq -n \
+      --arg run_id $RUN_ID \
+      --arg run_url  $RUN_URL \
+      --arg code_url  $CODE_URL \
+      --arg reason $(Build.Reason) \
+      --arg source_branch $(Build.SourceBranchName) \
+      --arg requester "$(Build.RequestedFor)" \
+      --arg scenario_name "$SCENARIO_NAME" \
```
alyssa1303, line 23: Can we collect GitHub branch and commit too for ADO runs? I believe this will only collect ADO branch and commit if for all internal pipelines

alyssa1303: It'll be useful if an external to track if external change break internal pipelines and how often it happen to improve our monitoring

sumanthreddy29: Sure. I added the change. I will test and update the PR and the Teams thread.

alyssa1303 (approved):

alyssa1303 (changes requested):
Actual just saw 1 error "line 37: Build.CronSchedule.DisplayName: command not found". Is this resolved?

alyssa1303 (approved):

PR #390, opened 2024-11-14
Use Nitro-based instance type in AWS for SLO discovery

rafael-mendes-pereira:
Use Nitro instance in AWS to support Prefix delegation and prevent ec2 API throttling

### DIFF
--- a/scenarios/perf-eval/slo-servicediscovery/terraform-inputs/aws.tfvars
+++ b/scenarios/perf-eval/slo-servicediscovery/terraform-inputs/aws.tfvars
@@ -96,7 +96,7 @@ eks_config_list = [{
     {
       name           = "userpool0"
       ami_type       = "AL2_x86_64"
-      instance_types = ["m4.xlarge"]
+      instance_types = ["m5a.xlarge"]
       min_size       = 300
       max_size       = 300
       desired_size   = 300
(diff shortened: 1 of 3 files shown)


scenarios/perf-eval/slo-servicediscovery/terraform-inputs/aws.tfvars
```diff
@@ -96,7 +96,7 @@ eks_config_list = [{
     {
       name           = "userpool0"
       ami_type       = "AL2_x86_64"
-      instance_types = ["m4.xlarge"]
+      instance_types = ["m5a.xlarge"]
```
alyssa1303, line 99: Can we change AWS only first? It seems v5 is used by other team so quota is still constrained even with 4k. We might need to raise it to 10k or 20k to avoid lack of quota failure

scenarios/perf-eval/slo-servicediscovery/terraform-inputs/azure.tfvars
```diff
@@ -58,7 +58,7 @@ aks_config_list = [
         min_count            = 300
         max_count            = 300
         auto_scaling_enabled = true
-        vm_size              = "Standard_D4_v3"
+        vm_size              = "Standard_D4_v5"
```
alyssa1303, line 61: Can we change AWS only first? It seems v5 is used by other team so quota is still constrained even with 4k. We might need to raise it to 10k or 20k to avoid lack of quota failure?

alyssa1303 (approved):

PR #393, opened 2024-11-15
Set WARM_PREFIX_TARGET for Slo Svc Discovery equal to max Pods per Node

rafael-mendes-pereira:
Set WARM_PREFIX_TARGET for Slo Svc Discovery equal to max Pods per Node to prevent throttling for ec2 API AssignPrivateIpAddresses.

- Add parameter to EKS configuration to set the WARM_PREFIX_TARGET
- Add unit tests for VPC config (test it locally with: `terraform test`)

### DIFF
--- a/modules/terraform/aws/eks/main.tf
+++ b/modules/terraform/aws/eks/main.tf
@@ -26,7 +26,11 @@ locals {
           # Nodes must be AWS Nitro-based (see: https://docs.aws.amazon.com/ec2/latest/instancetypes/ec2-nitro-instances.html#nitro-instance-types)
           # Note: we've seen that it also prevents ENIs leak caused the issue: https://github.com/aws/amazon-vpc-cni-k8s/issues/608
           ENABLE_PREFIX_DELEGATION = "true"
-          WARM_PREFIX_TARGET       = "1"
+
+          # Should set either WARM_PREFIX_TARGET or both MINIMUM_IP_TARGET and WARM_IP_TARGET (see: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/prefix-and-ip-target.md)
+          WARM_PREFIX_TARGET = var.eks_config.vpc_cni_minimum_ip_target == "0" ? "1" : "0"
+          MINIMUM_IP_TARGET  = var.eks_config.vpc_cni_minimum_ip_target
+          WARM_IP_TARGET     = var.eks_config.vpc_cni_minimum_ip_target != "0" ? "1" : "0"
 
           ADDITIONAL_ENI_TAGS = jsonencode(var.tags)
         }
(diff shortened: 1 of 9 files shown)


modules/terraform/aws/eks/main.tf
```diff
@@ -26,7 +26,11 @@ locals {
           # Nodes must be AWS Nitro-based (see: https://docs.aws.amazon.com/ec2/latest/instancetypes/ec2-nitro-instances.html#nitro-instance-types)
           # Note: we've seen that it also prevents ENIs leak caused the issue: https://github.com/aws/amazon-vpc-cni-k8s/issues/608
           ENABLE_PREFIX_DELEGATION = "true"
-          WARM_PREFIX_TARGET       = "1"
+
+          # Should set either WARM_PREFIX_TARGET or both MINIMUM_IP_TARGET and WARM_IP_TARGET (see: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/prefix-and-ip-target.md)
+          WARM_PREFIX_TARGET = var.eks_config.vpc_cni_minimum_ip_target == "0" ? "1" : "0"
```
alyssa1303, line 31: Why check for zero? Can we just check if the config exist or not? If not use default value set by vpc-cni?

modules/terraform/aws/eks/main.tf
```diff
@@ -26,7 +26,11 @@ locals {
...
+          MINIMUM_IP_TARGET  = var.eks_config.vpc_cni_minimum_ip_target
```
alyssa1303, line 32: Same here what if people don't set it?

modules/terraform/aws/eks/main.tf
```diff
@@ -26,7 +26,11 @@ locals {
...
+          WARM_IP_TARGET     = var.eks_config.vpc_cni_minimum_ip_target != "0" ? "1" : "0"
```
alyssa1303, line 33: Here too

modules/terraform/aws/eks/variables.tf
```diff
@@ -52,7 +52,8 @@ variable "eks_config" {
         env = optional(map(string))
       }))
     }))
-    kubernetes_version = optional(string, null)
+    vpc_cni_minimum_ip_target = optional(string, "0")
```
alyssa1303, line 55: I'd suggest the other 2 variables to be optional too and use default value defined by vpc-cni instead of 0. What if people want to change those values? We want prefix delegation to always enable but other values should be open for tuning

scenarios/perf-eval/slo-servicediscovery/terraform-inputs/aws.tfvars
```diff
@@ -146,6 +146,9 @@ eks_config_list = [{
     }
   ]
 
+  # Set IP target to max number of Pods per Node (58 for m5a.xlarge) + 2 for extras
+  vpc_cni_minimum_ip_target = "60"
```
alyssa1303, line 150: Please make this part of eks_addons config. This won't make sense if people don't even enable vpc_cni. Note that there are internal pipelines using aks module so we have to be careful

modules/terraform/aws/eks/variables.tf
rafael-mendes-pereira: I've replaced it with only `vpc_cni_warm_prefix_target`

alyssa1303 (approved):

PR #399, opened 2024-11-21
Add launch template resource to EKS node groups with default tags

rafael-mendes-pereira:
The goal of this change is to add the required tags (i.e. owner and deletion_due_time) to ec2 instances created by the EKS cluster.

Note: AWS does not recommend to update the default launch template created by EKS when no custom launch template is provided: 

> When you don’t provide a launch template, the Amazon EKS API creates one automatically with default values in your account. However, we don’t recommend that you modify auto-generated launch templates. Furthermore, existing node groups that don’t use a custom launch template can’t be updated directly. Instead, you must create a new node group with a custom launch template to do so

https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html

### DIFF
--- a/modules/terraform/aws/eks/main.tf
+++ b/modules/terraform/aws/eks/main.tf
@@ -158,7 +158,7 @@ resource "aws_ec2_tag" "cluster_security_group" {
 resource "aws_launch_template" "launch_template" {
   for_each = local.eks_node_group_map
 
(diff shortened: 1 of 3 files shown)


sumanthreddy29 (approved):

alyssa1303 (commented):
Please trigger a run for E2E test

alyssa1303 (approved):

PR #425, opened 2024-12-09
feat: setup recurring runs for cilium nodesubnet

santhoshmprabhu:
This PR adds 2 pipelines:
1. cilium-cluster-churn-cilium-nodesubnet runs the cluster churn scenario twice a day
2. slo-servicediscovery-cilium-nodesubnet runs the service churn scenario twice a day

These pipelines will use custom BYOCNI clusters because cilium nodesubnet is not available on AKS


jshr-w (approved):
LGTM, thanks!

PR #426, opened 2024-12-10
Append only RUN_ID to result file and update container to system for metatdata

alyssa1303:
Instead of adding all Telescope metadata, just add RUN_ID to cross check with metadata table to reduce duplicate

### DIFF
--- a/steps/collect-telescope-metadata.yml
+++ b/steps/collect-telescope-metadata.yml
@@ -73,16 +73,16 @@ steps:
     REGIONS: ${{ convertToJson(parameters.regions) }}
 
 - script: |
-    # Add telescope metadata to each JSON record in the test results file if it exists
+    # Append run_id to the test results file if the file exists
     set -eux
     if [ -f "$(TEST_RESULTS_FILE)" ]; then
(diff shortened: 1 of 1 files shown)


anson627 (approved):

steps/collect-telescope-metadata.yml
```diff
@@ -73,16 +73,16 @@ steps:
     REGIONS: ${{ convertToJson(parameters.regions) }}
 
 - script: |
-    # Add telescope metadata to each JSON record in the test results file if it exists
+    # Append run_id to the test results file if the file exists
     set -eux
     if [ -f "$(TEST_RESULTS_FILE)" ]; then
-      jq --argfile telescope_metadata $(TELESCOPE_METADATA_FILE) \
+      jq --arg telescope_metadata $(jq -n -c --arg run_id $RUN_ID '{run_id: $run_id}') \
```
anson627, line 79: just want to make sure it can handle the case where run_id field already exists

alyssa1303: We will still append even if run id exists to even consistent format for querying

anson627 (approved):

PR #427, opened 2024-12-10
fix kperf image

anson627:


pipelines/perf-eval/API Server Benchmark/apiserver-benchmark-virtualnodes10-pods100.yml
```diff
@@ -33,7 +33,7 @@ stages:
               extra_benchmark_subcmd_args: ""
               disable_warmup: "true"
           engine_input:
-            runner_image: ghcr.io/azure/kperf:v0.1.5
+            runner_image: ghcr.io/azure/kperf:0.1.5
```
alyssa1303, line 36: Is there a reason why some use 0.1.5 why other use 0.1.6? Should we use ghcr.io/azure/kperf:0.1.6 for all?

alyssa1303 (approved):

pipelines/perf-eval/API Server Benchmark/apiserver-benchmark-virtualnodes10-pods100.yml
anson627: let's fix the build first

PR #430, opened 2024-12-11
Add cilium overlay stage for comparison

santhoshmprabhu:
This PR adds a cilium overlay cluster to the cilium nodesubnet service and cluster churn pipelines for comparison.

The configuration for the added stages matches those in cilium service and cluster churn pipelines .


jshr-w (approved):
lgtm, after syntax fix

jshr-w (approved):

PR #431, opened 2024-12-12
Network churn

agrawaliti:


agrawaliti:
@microsoft-github-policy-service agree [company="Microsoft"]

agrawaliti:

steps/topology/network-churn/validate-resources.yml
```diff
@@ -0,0 +1,19 @@
...
+    type: string
+  - name: regions
+    type: object
+
+steps:
+  - template: /steps/cloud/${{ parameters.cloud }}/update-kubeconfig.yml
+    parameters:
+      role: net
+      region: ${{ parameters.regions[0] }}
+  - template: /steps/engine/clusterloader2/cilium/scale-cluster.yml
```
anson627, line 14: can we do this in terraform when setting up cluster?

steps/setup-tests.yml
```diff
@@ -9,27 +9,40 @@ parameters:
 - name: run_id
   type: string
   default: ''
+- name: run_id_2
```
anson627, line 12: same here

steps/setup-tests.yml
```diff
@@ -9,27 +9,40 @@ parameters:
...
+  type: string
+  default: ''
 - name: retry_attempt_count
   type: number
   default: 3
 - name: credential_type
   type: string
 - name: ssh_key_enabled
   type: boolean
+- name: use_secondary_cluster
```
anson627, line 22: same here

modules/python/clusterloader2/slo/config/load-config.yaml
```diff
@@ -29,17 +30,17 @@ name: load-config
 
 # Service test
 {{$BIG_GROUP_SIZE := DefaultParam .BIG_GROUP_SIZE 4000}}
-{{$SMALL_GROUP_SIZE := DefaultParam .SMALL_GROUP_SIZE 20}}
+{{$SMALL_GROUP_SIZE := DefaultParam .CL2_DEPLOYMENT_SIZE 20}}
```
jshr-w, line 33: Can we name this CL2_SMALL_GROUP_SIZE to keep the variable naming coordinated?

modules/python/clusterloader2/slo/config/load-config.yaml
```diff
@@ -29,17 +30,17 @@ name: load-config
...
 {{$bigDeploymentsPerNamespace := DefaultParam .bigDeploymentsPerNamespace 1}}
-{{$smallDeploymentPods := SubtractInt $podsPerNamespace (MultiplyInt $bigDeploymentsPerNamespace $BIG_GROUP_SIZE)}}
+{{$smallDeploymentPods :=  DivideInt $totalPods $namespaces}}
```
jshr-w, line 35: This is going to break all the other tests, right? Could you please restore this, and probably create a parameter for bigDeployments and set that to 0 instead.

modules/python/clusterloader2/slo/config/load-config.yaml
```diff
@@ -29,17 +30,17 @@ name: load-config
...
 {{$smallDeploymentsPerNamespace := DivideInt $smallDeploymentPods $SMALL_GROUP_SIZE}}
 
 namespace:
   number: {{$namespaces}}
   prefix: slo
   deleteStaleNamespaces: true
   deleteAutomanagedNamespaces: true
-  enableExistingNamespaces: false
+  enableExistingNamespaces: true
```
jshr-w, line 43: This may break testing. If namespaces weren't deleted by the previous run, we should be aware (many existing pipelines are dependent on this). Let's restore to the original value.

modules/python/clusterloader2/slo/config/modules/reconcile-objects.yaml
```diff
@@ -41,10 +49,11 @@ steps:
     - basename: big-deployment
       objectTemplatePath: deployment_template.yaml
       templateFillMap:
-        Replicas: {{$bigDeploymentSize}}
+        Replicas: {{$bigDeploymentSize}}kube
```
jshr-w, line 52: Is this a typo?

modules/python/clusterloader2/slo/config/modules/reconcile-objects.yaml
```diff
@@ -41,10 +49,11 @@ steps:
...
         SvcName: big-service
         Group: {{.Group}}
         deploymentLabel: {{.deploymentLabel}}
+{{end}}
   - namespaceRange:
```
jshr-w, line 57: We don't want this code to execute if we are not running a network policy right? Shouldn't we 'else' gate this?

modules/python/clusterloader2/slo/slo.py
```diff
@@ -24,13 +24,9 @@
 }
 # TODO: Remove aks once CL2 update provider name to be azure
 
-def calculate_config(cpu_per_node, node_count, provider, service_test):
+def calculate_config(cpu_per_node, node_per_step, pods_per_node, provider):
     throughput = 100
-    nodes_per_namespace = min(node_count, DEFAULT_NODES_PER_NAMESPACE)
```
jshr-w, line 29: I think the changes made to pods_per_node are going to break many of the existing pipelines. After this change, how can the service test use 20 pods per node? We need to be careful adding parameters given the number of pipelines that are dependent on them... IMO the safest way will be to have an IF branch here, and possibly a parameter for pods per node ONLY for the network_test.

modules/python/clusterloader2/slo/slo.py
```diff
@@ -142,7 +167,7 @@ def collect_clusterloader2(
         "group": None,
         "measurement": None,
         "result": None,
-        # "test_details": details,
+        # # "test_details": details,
```
jshr-w, line 170: nit: typo

modules/python/clusterloader2/slo/slo.py
```diff
@@ -192,13 +217,18 @@ def main():
     parser_configure.add_argument("node_count", type=int, help="Number of nodes")
     parser_configure.add_argument("node_per_step", type=int, help="Number of nodes per scaling step")
     parser_configure.add_argument("max_pods", type=int, help="Maximum number of pods per node")
+    parser_configure.add_argument("pods_per_node", type=int, help="Number of pods per node")
     parser_configure.add_argument("repeats", type=int, help="Number of times to repeat the deployment churn")
     parser_configure.add_argument("operation_timeout", type=str, help="Timeout before failing the scale up test")
+    parser_configure.add_argument("no_of_namespaces", type=int, default=1, help="Number of namespaces to create")
```
jshr-w, line 223: adding arguments without a default forced (need to set `nargs`) will probably break all the other pipelines in my understand... this comment applies to all arguments added

steps/engine/clusterloader2/cilium/scale-cluster.yml
```diff
@@ -43,8 +43,8 @@ steps:
       if [ "true" =  "${{ parameters.enable_autoscale }}" ]; then
          az aks nodepool update --cluster-name $aks_name --name $np --resource-group $aks_rg --enable-cluster-autoscaler --min-count 0 --max-count 500
       fi
-      az aks nodepool update --cluster-name $aks_name --name $np --resource-group $aks_rg --node-taints "slo=true:NoSchedule" --labels slo=true
-      sleep 300
+      az aks nodepool update --cluster-name $aks_name --name $np --resource-group $aks_rg --labels slo=true test-np=net-policy-client
+      # sleep 300
```
jshr-w, line 47: This is going to affect all the other pipelines... please let's be careful!

jshr-w (changes requested):
Let's try to (1) Minimize changes that touch code that other pipelines use (2) After minimizing those changes, need to run the other automated pipelines off this branch to ensure they aren't broken.

steps/engine/clusterloader2/cilium/scale-cluster.yml
agrawaliti: Hello I pushed some test commits yesterday, I am cleaning it up. Thanks for pointing out

jobs/competitive-test.yml
```diff
@@ -33,6 +33,9 @@ parameters:
 - name: run_id
   type: string
   default: ''
+- name: run_id_2
```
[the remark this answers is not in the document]
agrawaliti: I am using two different precreated cluster for azure_cilium and azure_cni_overlay and i am passing those two clusters using run_id and run_id_2, but on second thought i am thinking we can create clusters using terraform too, that will remove this parameter. I'll move cluster creation in terraform and remove this.

modules/python/clusterloader2/slo/config/modules/reconcile-objects.yaml
agrawaliti: I dont wanna run bigDeployment for network test so i have added {{if not $NETWORK_TEST}} for big deployment

modules/python/clusterloader2/slo/slo.py
agrawaliti: in my opinion having pods_per_node as a hard constant which can change frequently based on usecase in not a good approach, I have added that parameter in pipeline configuration so if we need custom value we can set it in pipeline, else if unset it will keep working as before i.e default - 40

modules/python/clusterloader2/slo/config/modules/reconcile-objects.yaml
jshr-w: I mean, it works for your scenario but it will break the others -- If `not NETWORK_TEST`, what is stopping the pipeline from running both phases?

jshr-w (changes requested):

PR #440, opened 2024-12-18
Add network_dataplane and network_policy as variables for AKS module

rafael-mendes-pereira:
- Update AKS module to support custom data plane configuration
- Add unit tests for network_dataplane and network_policy variables
- Add validation for AKS network policy and data plane compatibility

### DIFF
--- a/modules/terraform/azure/main.tf
+++ b/modules/terraform/azure/main.tf
@@ -32,7 +32,7 @@ locals {
           aks.network_profile,
           {
             network_policy    = local.aks_network_policy != null ? local.aks_network_policy : aks.network_profile.network_policy
-            network_dataplane = local.aks_network_dataplane != null ? local.aks_network_dataplane : aks.network_profile.network_dataplane
+            network_dataplane = aks.network_profile.network_dataplane
           }
         )
       }
@@ -92,6 +92,7 @@ module "aks" {
   vnet_id             = try(module.virtual_network[each.value.role].vnet_id, null)
   subnets             = try(local.all_subnets, null)
   k8s_machine_type    = local.k8s_machine_type
+  network_dataplane   = local.aks_network_dataplane
 }
 
 module "aks-cli" {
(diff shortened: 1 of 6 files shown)


pipelines/perf-eval/CNI Benchmark/cilium-cluster-churn-nodes1000-pods50k.yml
```diff
@@ -1,45 +1,45 @@
 trigger: none
-schedules:
-  - cron: "0 */12 * * *"
-    displayName: "Every 12 Hour"
-    branches:
-      include:
-        - main
-    always: true
+# schedules:
```
alyssa1303, line 2: revert change

modules/terraform/azure/tests/test_aks_network_dataplane.tftest.hcl
```diff
@@ -0,0 +1,120 @@
...
+      role        = "test"
+      aks_name    = "test"
+      dns_prefix  = "test"
+      subnet_name = "test-subnet-1"
+      sku_tier    = "Standard"
+      network_profile = {
+        network_plugin      = "azure"
+        network_plugin_mode = "overlay"
+        network_policy      = "azure"
+        network_dataplane   = "azure"
```
alyssa1303, line 25: Can we add test case for when `network_policy` and `network_dataplane` aren't set and both variables from pipelines aren't set either?

rafael-mendes-pereira: That is only possible for `network_dataplane` since `network_policy` is defined, when not set, only during the provisioning phase (apply).

modules/terraform/azure/main.tf
```diff
@@ -28,13 +28,7 @@ locals {
...
         sku_tier           = local.aks_sku_tier != null ? local.aks_sku_tier : aks.sku_tier
         kubernetes_version = local.aks_kubernetes_version != null ? local.aks_kubernetes_version : aks.kubernetes_version
-        network_profile = merge(
-          aks.network_profile,
-          {
-            network_policy    = local.aks_network_policy != null ? local.aks_network_policy : aks.network_profile.network_policy
-            network_dataplane = local.aks_network_dataplane != null ? local.aks_network_dataplane : aks.network_profile.network_dataplane
-          }
-        )
+        network_profile    = aks.network_profile
```
alyssa1303, line 31: I believe this is not needed anymore if we don't override here?

alyssa1303 (approved):

PR #447, opened 2024-12-20
Add optional subnet_names to EKS node group configuration

rafael-mendes-pereira:
This allows to define, for example, that all nodes in a group would be in the availability zones of a specific region.

Note: we cannot create unit tests for it since the subnet ids are defined only after the resource is created.


.github/workflows/terraform-validation.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Setup Terraform
         uses: hashicorp/setup-terraform@v2
         with:
-          terraform_version: 1.6.0
+          terraform_version: 1.7.0
```
alyssa1303, line 26: let's update to 1.10 since it's the version used in pipelines

modules/terraform/aws/eks/main.tf
```diff
@@ -59,6 +61,21 @@ data "aws_subnets" "subnets" {
   }
 }
 
+data "aws_subnet" "subnet" {
```
alyssa1303, line 64: Should call this subnets_by_names to distinguish from the subnets data above (which is subnets by vpc)

modules/terraform/aws/eks/variables.tf
```diff
@@ -43,6 +43,7 @@ variable "eks_config" {
       desired_size   = number
       capacity_type  = optional(string, "ON_DEMAND")
       labels         = optional(map(string), {})
+      subnet_ids     = optional(list(string), null)
```
alyssa1303, line 46: This should be subnet_names. Subnet ID is different

modules/terraform/aws/variables.tf
```diff
@@ -111,6 +111,7 @@ variable "eks_config_list" {
       desired_size   = number
       capacity_type  = optional(string, "ON_DEMAND")
       labels         = optional(map(string), {})
+      subnet_ids     = optional(list(string), null)
```
alyssa1303, line 114: Rename to subnet_names

modules/terraform/aws/eks/main.tf
rafael-mendes-pereira: They are different types: 
-  [aws_subnets](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/subnets) gets the subnets ids only
-  [aws_subnet](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/subnet) gets the subnet details.

rafael-mendes-pereira: I've updated it to `aws_subnet.subnet_details`

modules/terraform/aws/eks/variables.tf
rafael-mendes-pereira: Good catch. Thanks

alyssa1303 (approved):

PR #451, opened 2024-12-23
Create 40k pods in the cluster churn pipeline

santhoshmprabhu:
This PR lowers the number of pods per node in the cluster churn scenarios to 40. In addition, we also remove the second stages from the cilium nodesubnet service/cluster churn pipelines. The second change has no real impact, since the pipelines were already using the PR branch.


jshr-w (approved):
thanks for the change!

PR #488, opened 2025-01-24
*: update k8s version and track extra information

fuweid:


steps/engine/kperf/collect.yml
```diff
@@ -45,11 +45,13 @@ steps:
...
     fi
 
+    k8s_version="$(echo $CLOUD_INFO | jq -r .k8s_version)"
+
     cat ${TEST_RESULTS_DIR}/tmp_results.json \
       | jq 'del(.result.errors)' \
       | jq -c ".info += { flowcontrol: \"$FLOWCONTROL\", subcmdArgs: \"$BENCHMARK_SUBCMD_ARGS\", extraArgs: \"$EXTRA_BENCHMARK_SUBCMD_ARGS\" }" \
       | jq -c ". += { timestamp: $(date -u +\"%Y-%m-%dT%H:%M:%SZ\") }" \
-      | jq -c ". += { runID: \"$RUN_ID\", runURL: \"$RUN_URL\", cloud: \"$CLOUD_NAME\" }" > "${TEST_RESULTS_FILE}"
+      | jq -c ". += { runID: \"$RUN_ID\", runURL: \"$RUN_URL\", cloud: \"$CLOUD_NAME\", k8s_version: \"$k8s_version\" }" > "${TEST_RESULTS_FILE}"
```
alyssa1303, line 54: Is there a reason why we cannot directly use CLOUD_INFO?

scenarios/perf-eval/apiserver-vn100pod10k/terraform-inputs/azure.tfvars
```diff
@@ -5,9 +5,10 @@ owner          = "aks"
 
 aks_cli_config_list = [
   {
-    role     = "client"
-    aks_name = "vn100-p10k"
-    sku_tier = "standard"
+    role               = "client"
+    aks_name           = "vn100-p10k"
+    sku_tier           = "standard"
+    kubernetes_version = "1.31"
```
alyssa1303, line 11: Can we do the same thing for AWS so we can compare same version? If not specified, AWS will take latest version which might be 1.32 later while Azure is still 1.31

steps/engine/kperf/collect.yml
fuweid: good point. I was thinking that it's better to limit the number of exported fields.
let me update it

scenarios/perf-eval/apiserver-vn100pod10k/terraform-inputs/azure.tfvars
fuweid: sure. updated

alyssa1303 (approved):

sumanthreddy29 (approved):

PR #495, opened 2025-01-28
Break down CNI pipelines into separate schedules for Azure and AWS

alyssa1303:
- Use separate schedules for AWS and Azure
- Add a variable `SKIP_AWS` to skip AWS stage by default for manual run


pipelines/perf-eval/CNI Benchmark/cilium-cluster-churn-nodes1000-pods50k-azure.yml
```diff
@@ -1,7 +1,7 @@
 trigger: none
 schedules:
   - cron: "0 */12 * * *"
-    displayName: "Every 12 Hour"
+    displayName: "12:00 AM & PM Daily"
```
sumanthreddy29, line 4: Can you try this with one pipeline and multiple schedules as part of same pipeline?

https://learn.microsoft.com/en-us/azure/devops/pipelines/process/scheduled-triggers?view=azure-devops&tabs=yaml#buildcronscheduledisplayname-variable

 I tested this for CRUD Intv2 and stag before and it worked.

sumanthreddy29:
```
- stage: stage1
  # Run this stage only when the pipeline is triggered by the 
  # "Daily midnight build" cron schedule
  condition: eq(variables['Build.CronSchedule.DisplayName'], 'Daily midnight build')
```

sumanthreddy29 (changes requested):

pipelines/perf-eval/CNI Benchmark/cilium-cluster-churn-nodes1000-pods50k-azure.yml
alyssa1303: done

sumanthreddy29 (approved):

jshr-w (approved):
Thanks Alyssa!

PR #500, opened 2025-01-29
Additions to CCNP/CNP Pipeline

karina-ranadive:
Hello, 

I had a few cilium metrics for measurements for ccnp & cnp scale testing. I also adjusted the cnp and ccnp policies being deployed. I also made it possible for the user to enter a number for repeats. I also changed the stage name.


steps/engine/clusterloader2/slo/collect.yml
karina-ranadive, line 1: I changed $REPEATS to ${REPEATS:-1} just in case in pipeline user does not input a value for repeats just so there are no errors in case

karina-ranadive:
going to remove this pipeline eventually - merging into network policy scale pipeline

PR #506, opened 2025-02-03
Update Auoscaler Module and tests scenarios

sumanthreddy29:
1. Add support for warmup deployment for NAP to calculate the available CPU request
2. Update CAS tests to pre-create a user node and use it to calculate the CPU request.
3. Update Config.yaml to take desiredNodecount parameters instead of scaling down to 0
4. Updated all CAS tests with autoscaler profile.
5. Added k8s version for NAP tests.


scenarios/perf-eval/cas-c2n200p200/terraform-inputs/azure.tfvars
```diff
@@ -37,8 +37,15 @@ aks_config_list = [
     ]
     kubernetes_version = "1.31"
     auto_scaler_profile = {
-      scale_down_delay_after_add = "0m"
-      scale_down_unneeded        = "0m"
+      scale_down_delay_after_add     = "2m"
```
alyssa1303, line 40: can we set the same profile on AWS so we have fair comparison?

modules/python/clusterloader2/autoscale/autoscale.py
```diff
@@ -7,8 +7,8 @@
 from utils import parse_xml_to_json, run_cl2_command
 
 def override_config_clusterloader2(cpu_per_node, node_count, pod_count, scale_up_timeout, scale_down_timeout, loop_count, node_label_selector, node_selector, override_file):
-    # assuming 85% of the CPU cores can be used by test pods
-    cpu_request = (cpu_per_node * 1000 * 0.85) * node_count // pod_count
+    # assuming 75% of the CPU cores can be used by test pods
```
alyssa1303, line 10: This percentage is very sensitive to vm size, so let leave a note here regarding which vm size you're looking at. Also, we might need different values for different clouds since each cloud reserve differently

scenarios/perf-eval/cas-c2n200p200/terraform-inputs/azure.tfvars
sumanthreddy29: Sure. Updated for aws.

modules/python/clusterloader2/autoscale/autoscale.py
sumanthreddy29: I updated the logic to keep 85% for AWS and 75% for Azure.

modules/python/clusterloader2/autoscale/autoscale.py
```diff
@@ -145,7 +151,7 @@ def main():
     args = parser.parse_args()
 
     if args.command == "override":
-        override_config_clusterloader2(args.cpu_per_node, args.node_count, args.pod_count, args.scale_up_timeout, args.scale_down_timeout, args.loop_count, args.node_label_selector, args.node_selector, args.cl2_override_file)
+        override_config_clusterloader2(args.provider, args.cpu_per_node, args.node_count, args.pod_count, args.scale_up_timeout, args.scale_down_timeout, args.loop_count, args.node_label_selector, args.node_selector, args.cl2_override_file)
```
alyssa1303, line 154: if you don't update yaml file, this will break

sumanthreddy29: updated

modules/python/clusterloader2/autoscale/autoscale.py
```diff
@@ -6,9 +6,15 @@
 from datetime import datetime, timezone
 from utils import parse_xml_to_json, run_cl2_command
 
-def override_config_clusterloader2(cpu_per_node, node_count, pod_count, scale_up_timeout, scale_down_timeout, loop_count, node_label_selector, node_selector, override_file):
-    # assuming 85% of the CPU cores can be used by test pods
-    cpu_request = (cpu_per_node * 1000 * 0.85) * node_count // pod_count
+def override_config_clusterloader2(provider, cpu_per_node, node_count, pod_count, scale_up_timeout, scale_down_timeout, loop_count, node_label_selector, node_selector, override_file):    
+    if provider == "aks":
```
anson627, line 10: let's set 0.75 on both side to be fair

alyssa1303 (approved):

anson627 (dismissed):

steps/engine/clusterloader2/autoscale/execute.yml
```diff
@@ -11,10 +11,14 @@ parameters:
...
     set -eo pipefail
+    if [ -z "$WARMUP_DEPLOYMENT" ]; then
+      WARMUP_DEPLOYMENT="false"
+    fi
+
     PYTHONPATH=$PYTHONPATH:$(pwd) python3 $PYTHON_SCRIPT_FILE override \
       $CPU_PER_NODE $NODE_COUNT $POD_COUNT \
       $SCALE_UP_TIMEOUT $SCALE_DOWN_TIMEOUT \
-      $LOOP_COUNT "$NODE_LABEL_SELECTOR" "$NODE_SELECTOR" ${CL2_CONFIG_DIR}/overrides.yaml
+      $LOOP_COUNT "$NODE_LABEL_SELECTOR" "$NODE_SELECTOR" ${CL2_CONFIG_DIR}/overrides.yaml $WARMUP_DEPLOYMENT
```
alyssa1303, line 21: We can do this to remove the if statement `${WARMUP_DEPLOYMENT:-false}`

sumanthreddy29: Updated

alyssa1303 (approved):

PR #512, opened 2025-02-05
Use ACR for pods deployed in Azure

alyssa1303:
Use ACR for pods deployed in Azure


sumanthreddy29 (approved):

sumanthreddy29 (approved):

modules/python/clusterloader2/cri/cri.py
```diff
@@ -75,6 +75,7 @@ def override_config_clusterloader2(
         file.write("CL2_PROMETHEUS_MEMORY_SCALE_FACTOR: 30.0\n")
         file.write("CL2_PROMETHEUS_NODE_SELECTOR: \"prometheus: \\\"true\\\"\"\n")
         file.write(f"CL2_POD_STARTUP_LATENCY_THRESHOLD: {pod_startup_latency_threshold}\n")
+        file.write(f"CL2_PROVIDER: {provider}\n")
```
sumanthreddy29, line 78: add try-catch block for k8s.client operations

alyssa1303: Can add later if it's really needed. In case of api server failure, test shouldn't proceed so exception raise is expected

PR #516, opened 2025-02-06
add measurements for cl2

jshr-w:
This PR makes some miscellaneous changes to the Cilium metrics in preparation for some feature testing.
- adds some measurements to the SLO CL2 tests
- We also move all the service churn tests to use `max_pods` as the number of pods per node deployed, as part of the transition process to allow flexibility. 
- This PR also parameterizes some inputs from the service churn feature test pipeline to allow customizability.
- It also captures the start time for tests by default, to prevent inaccuracies in the dashboard due to startTime being set to time range filter start time for some queries.

### DIFF
--- a/modules/python/clusterloader2/slo/config/modules/measurements.yaml
+++ b/modules/python/clusterloader2/slo/config/modules/measurements.yaml
@@ -47,36 +47,6 @@ steps:
           pingSleepDuration: {{$PROBE_MEASUREMENTS_PING_SLEEP_DURATION}}
           threshold: {{$NETWORK_LATENCY_THRESHOLD}}
     {{end}}
-      - Identifier: NodeAvgCPUUsage
-        Method: GenericPrometheusQuery
-        Params:
-          action: {{$action}}
-          metricName: Node Average CPU Usage
-          metricVersion: v1
-          unit: cpu
-          enableViolations: true
-          queries:
-          - name: Perc99
-            query: quantile(0.99, avg_over_time(rate(node_cpu_usage_seconds_total[1m])[%v:]))
-          - name: Perc90
-            query: quantile(0.90, avg_over_time(rate(node_cpu_usage_seconds_total[1m])[%v:]))
-          - name: Perc50
-            query: quantile(0.50, avg_over_time(rate(node_cpu_usage_seconds_total[1m])[%v:]))
-      - Identifier: NodeMaxMemUsage
-        Method: GenericPrometheusQuery
-        Params:
-          action: {{$action}}
-          metricName: Node Max Memory Usage
-          metricVersion: v1
-          unit: MB
-          enableViolations: true
-          queries:
-          - name: Perc99
-            query: quantile(0.99, max_over_time(node_memory_working_set_bytes[%v:]) / 1024 / 1024)
-          - name: Perc90
-            query: quantile(0.90, max_over_time(node_memory_working_set_bytes[%v:]) / 1024 / 1024)
-          - name: Perc50
-            query: quantile(0.5, max_over_time(node_memory_working_set_bytes[%v:]) / 1024 / 1024)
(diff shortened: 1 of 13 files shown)


modules/python/clusterloader2/slo/config/modules/measurements.yaml
```diff
@@ -46,63 +47,17 @@ steps:
           pingSleepDuration: {{$PROBE_MEASUREMENTS_PING_SLEEP_DURATION}}
           threshold: {{$NETWORK_LATENCY_THRESHOLD}}
     {{end}}
-      - Identifier: ApiserverAvgCPUUsage
```
alyssa1303, line 49: Why do we remove these metrics?

modules/python/clusterloader2/slo/slo.py
```diff
@@ -30,7 +30,7 @@ def calculate_config(cpu_per_node, node_count, max_pods, provider, service_test,
 
     pods_per_node = DEFAULT_PODS_PER_NODE
     if service_test:
-        pods_per_node = LOAD_PODS_PER_NODE
```
alyssa1303, line 33: Remove this variable defined in line 11

modules/python/clusterloader2/slo/slo.py
```diff
@@ -79,6 +80,7 @@ def configure_clusterloader2(
         file.write("CL2_PROMETHEUS_MEMORY_SCALE_FACTOR: 30.0\n")
         file.write("CL2_PROMETHEUS_NODE_SELECTOR: \"prometheus: \\\"true\\\"\"\n")
         file.write("CL2_POD_STARTUP_LATENCY_THRESHOLD: 3m\n")
+        file.write("CL2_ENABLE_TERMINATED_WATCHES_MEASUREMENT: true\n")
```
alyssa1303, line 83: If it's always enabled, do we need to make a variable for it in the config? We can just add the measurement for all cases right?

steps/engine/clusterloader2/slo/execute.yml
```diff
@@ -9,6 +9,13 @@ parameters:
     type: string
 
 steps:
+  - script: |
+      echo "Set the start time for test execution"
+      startTimestamp=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
+      echo "Start: $startTimestamp"
+      echo "##vso[task.setvariable variable=START]$startTimestamp"
```
alyssa1303, line 16: Can we use a better name? Like SLO_START_TIME?

alyssa1303 (commented):
There is a new [pipeline](https://github.com/Azure/telescope/blob/main/pipelines/perf-eval/Scheduler%20Benchmark/pod-churn-50k-sched.yml#L35) that's also run this test so you can help update the max_pod to 20 as well

modules/python/clusterloader2/slo/config/modules/measurements.yaml
jshr-w: Unused in dashboards

modules/python/clusterloader2/slo/slo.py
jshr-w: Prefer to align with upstream. Defaulted true

modules/python/clusterloader2/slo/config/modules/measurements.yaml
jshr-w: added it back to avoid dealing with breaking anyone, though these shouldn't be in use.

alyssa1303 (approved):

modules/python/clusterloader2/slo/slo.py
alyssa1303: My question is more about since you don't really allow customization from pipeline at the execute step, then the value will always be True, so there isn't really a need to make a variable for this.

PR #540, opened 2025-02-21
feat: cilium network load test pipeline

huntergregory:
## Goal

Validate scenario with induce heavy pod traffic on many nodes.

## Validations

We query for hubble metrics and send to storage account.

### Future Work

Test the following:

1. cilium pod is ok.
2. Datapath is ok (e.g. conntrack).
3. Network observability (Hubble) metrics are ok.

## Topology

Topology when nodes=2, serversPerNode=2, clientsPerNode=2, namespaces=4, deployments=2.

**You can specify queries per second coming from each client as well.**

image

### Topology Parameters

- _nodes_
- _servers per node_
- _clients per node_
- _queries per second_ (per client)
- _namespaces_. There is **1 service per namespace**, and all server deployments in the namespace are its backend pods.
- _deployments_. A value of _N_ means _N_ client deployments in the cluster, each talking to a corresponding server deployment. **Controls the number of cilium identities.**

## Dependencies

Requires pending CL2 support for scraping hubble metrics.

### Other Parameters

- whether to apply a CNP that generates DNS metrics (requires ACNS security agent)


modules/python/clusterloader2/network-load/config/load-config.yaml
```diff
@@ -0,0 +1,155 @@
...
+      path: /modules/ciliumnetworkpolicy.yaml
+      params:
+        actionName: "Creating"
+        namespaces: {{$fortioNamespaces}}
+        Group: {{$groupName}}
+        cnpsPerNamespace: 1
+  {{end}}
+
+# deployment creation
+{{range $i := Loop $roundsToCreateOrRecreate}}
```
jshr-w, line 98: Wondering, is the goal here to only create deployments/servers/clients one time?

pipelines/perf-eval/Pod Traffic Benchmark/cilium-acns-network-load.yml
```diff
@@ -0,0 +1,50 @@
...
+      or(
+        eq(variables['Build.CronSchedule.DisplayName'], '6:00 AM & PM Daily'),
+        eq(variables['Build.Reason'], 'Manual')
+      )
+    dependsOn: []
+    jobs:
+      - template: /jobs/competitive-test.yml
+        parameters:
+          cloud: azure
+          regions:
```
jshr-w, line 27: Thinking out loud, you may want some of these parameters to be read from the pipeline instead for future flexibility?

modules/python/clusterloader2/network-load/config/load-config.yaml
huntergregory: We currently create all deployments at once. Then this variable allows you to do multiple rounds of deleting/recreating the deployments all at once

pipelines/perf-eval/Pod Traffic Benchmark/cilium-acns-network-load.yml
huntergregory: It looks like other perf evals under CNI benchmark use fixed parameters. Could be a good future enhancement for this and other evals

alyssa1303 (commented):
Can we consolidate network_load folder with slo folder by having a separate `network_load.py` as the entrypoint but re-using modules, measurements and functions whenever possible?

pipelines/perf-eval/Pod Traffic Benchmark/cilium-acns-network-load.yml
```diff
@@ -0,0 +1,50 @@
...
+              cpu_per_node: 4
+              node_count: 10
+              fortio_servers_per_node: 20
+              fortio_clients_per_node: 20
+              fortio_client_queries_per_second: 6300
+              fortio_client_connections: 6300
+              fortio_namespaces: 1
+              fortio_deployments_per_namespace: 1
+              apply_fqdn_cnp: True
+          max_parallel: 2
```
alyssa1303, line 47: There's only 1 job so this should be 1

pipelines/perf-eval/Pod Traffic Benchmark/cilium-acns-network-load.yml
```diff
@@ -0,0 +1,50 @@
...
+    dependsOn: []
+    jobs:
+      - template: /jobs/competitive-test.yml
+        parameters:
+          cloud: azure
+          regions:
+            - $(LOCATION)
+          engine: clusterloader2
+          engine_input:
+            image: "acnpublic.azurecr.io/clusterloader2:hgregory-02-26-test15-replace-address" # "ghcr.io/azure/clusterloader2:v20241022"
```
alyssa1303, line 31: What's the reason for using this image?

pipelines/perf-eval/Pod Traffic Benchmark/cilium-acns-network-load.yml
```diff
@@ -0,0 +1,50 @@
...
+    always: true
+
+variables:
+  SCENARIO_TYPE: perf-eval
+  SCENARIO_NAME: cilium-acns-network-load
+  OWNER: aks
+
+stages:
+  - stage: azure_westus2
+    condition: |
```
alyssa1303, line 17: This condition is not needed since you only have 1 stage

steps/topology/network-load/collect-clusterloader2.yml
```diff
@@ -0,0 +1,23 @@
...
+  parameters:
+    cloud: ${{ parameters.cloud }}
+    engine_input: ${{ parameters.engine_input }}
+    region: ${{ parameters.regions[0] }}
+
+- script: |
+    run_id=$(Build.BuildId)-$(System.JobId)
+    echo "Run ID: $run_id"
+    echo "##vso[task.setvariable variable=RUN_ID]$run_id"
+  displayName: "Set unique Run ID before publish"
```
alyssa1303, line 23: is there a reason why you don't use Terraform to provision cluster but use BYO cluster here?

huntergregory:
> Can we consolidate network_load folder with slo folder by having a separate `network_load.py` as the entrypoint but re-using modules, measurements and functions whenever possible? 

Thanks for the review. I originally tried this approach, but this data plane test is fundamentally different from the control plane tests in SLO. We would need to introduce more modules and many if-else statements and parameters specific to data plane testing. There would be convoluted logic to use/extend existing parameters to provide the new features and flexibility we're looking for in the PR description's topology.

Happy to talk about this more in depth and get your perspective.

pipelines/perf-eval/Pod Traffic Benchmark/cilium-acns-network-load.yml
huntergregory: This is a temporary workaround until another cluster loader 2 PR with Hubble metric scrape support is merged upstream

steps/topology/network-load/collect-clusterloader2.yml
huntergregory: We haven't built support for terraform yet. We'd prefer adding support in a follow up PR if that's ok

pipelines/perf-eval/Pod Traffic Benchmark/cilium-acns-network-load.yml
```diff
@@ -0,0 +1,50 @@
+trigger: none
+schedules:
```
alyssa1303, line 2: Can you do schedule after terraform implementation and update image?

pipelines/perf-eval/Pod Traffic Benchmark/cilium-acns-network-load.yml
alyssa1303: Then can you remove this condition here?

modules/python/clusterloader2/network-load/network_load.py
```diff
@@ -0,0 +1,266 @@
...
+    parser_configure.add_argument("--fortio-client-queries-per-second", type=int, required=True, help="Queries per second for each Fortio client pod. NOT queries per second per connection")
+    parser_configure.add_argument("--fortio-client-connections", type=int, required=True, help="Number of simultaneous connections for each Fortio client")
+    parser_configure.add_argument("--fortio-namespaces", type=int, required=True, help="Number of namespaces, each with their own service. Fortio clients query servers in the same namespace. Be weary of integer division causing less pods than expected regarding this parameter, pods, and pods per node.")
+    parser_configure.add_argument("--fortio-deployments-per-namespace", type=int, required=True, help="Number of Fortio server deployments (and number of client deployments) per service/partition. Be weary of integer division causing less pods than expected regarding this parameter, namespaces, pods, and pods per node.")
+    parser_configure.add_argument("--apply-fqdn-cnp", type=eval, choices=[True, False], default=False, help="Apply CNP that will generate DNS metrics")
```
MikeZappa87, line 198: I just merged this PR to remove the type=eval. It's a RCE, you can pass arbitrary code with this. 

https://github.com/Azure/telescope/pull/584

alyssa1303, line 198: We don't want to use type=eval due to some security issue, so please use this [function](https://github.com/Azure/telescope/blob/main/modules/python/clusterloader2/utils.py#L135) instead. Example [here](https://github.com/Azure/telescope/blob/main/modules/python/clusterloader2/slo/network_policy_scale.py#L211)

alyssa1303:
> > Can we consolidate network_load folder with slo folder by having a separate `network_load.py` as the entrypoint but re-using modules, measurements and functions whenever possible?
> 
> Thanks for the review. We are reusing slo validation functionality, and I originally tried reusing other pieces, but this data plane test is fundamentally different from the control plane tests in SLO. We would need to introduce more modules and many if-else statements and parameters specific to data plane testing. There would be convoluted logic to use/extend existing parameters to provide the new features and flexibility we're looking for in the PR description's topology.
> 
> Happy to talk about this more in depth and get your perspective.

Yes we also have the same discussion about that with Sarath. I think you can still have your own entrypoint using `network_load.py`, but there's still a way to re-use functions and modules like these PRs: https://github.com/Azure/telescope/pull/569 and https://github.com/Azure/telescope/pull/570

modules/python/clusterloader2/network-load/network_load.py
huntergregory: fixed

huntergregory: fixed

pipelines/perf-eval/Pod Traffic Benchmark/cilium-acns-network-load.yml
huntergregory: removed

pipelines/perf-eval/Pod Traffic Benchmark/cilium-acns-network-load.yml
huntergregory: yes, commented out the schedule

huntergregory:
> > > Can we consolidate network_load folder with slo folder by having a separate `network_load.py` as the entrypoint but re-using modules, measurements and functions whenever possible?
> > 
> > 
> > Thanks for the review. We are reusing slo validation functionality, and I originally tried reusing other pieces, but this data plane test is fundamentally different from the control plane tests in SLO. We would need to introduce more modules and many if-else statements and parameters specific to data plane testing. There would be convoluted logic to use/extend existing parameters to provide the new features and flexibility we're looking for in the PR description's topology.
> > Happy to talk about this more in depth and get your perspective.
> 
> Yes we also have the same discussion about that with Sarath. You can have your own entrypoint using `network_load.py`, but there's still a way to re-use functions and modules like these PRs: #569 and #570

The only module we could reuse seems to be [cilium-measurements.yaml](https://github.com/Azure/telescope/blob/main/modules/python/clusterloader2/slo/config/modules/cilium-measurements.yaml). Is there a way to reference files within _modules/python/clusterloader2/slo/modules/_ from _modules/python/clusterloader2/network-load/config/load-config.yaml_?

I ask because these data plane tests provide a new topology with its own versions of _service_template.yaml_, _cnp_template.yaml_, _deployment_template.yaml_, _modules/reconcile-objects.yaml_, _modules/services.yaml_, and _modules/ciliumnetworkpolicy.yaml_. If we move _.../network-load/config/_ to _.../slo/config/_ folder, then there could be confusion since the control plane tests will use a different subset of files than the data plane tests in the same _slo/_ folder.

alyssa1303 (approved):

PR #555, opened 2025-03-05
enable azure policy add on for Karpenter

PabloTriv:
Create the cluster with the azure policy enabled to prevent it from being enabled later on. Enabling the cluster mid-test triggers a control plane restart that adds a lot of latency during the tests.


PabloTriv:
image

PR #575, opened 2025-03-24
CCP Job Scheduling Latency Benchmark 

vittoriasalim:
This pull request introduces job scheduling latency benchmark using ClusterLoader2 (CL2) and KWOK, as a performance evaluation for job controller in Kubernetes.

- Added **`JobLifecycleLatency`** measurement in [`modules/python/clusterloader2/utils.py`]
- Added performance evaluation pipeline for job scheduling.
- Added cl2 config file for job scheduling performance test
- Added cl2 job template for job execution process.
- Added job scheduling evaluation scenarios. To provide environment-specific inputs for job-scheduling tests on AWS and Azure.
- Introduced mock-node KWOK topology for simulating job workloads
- Define CL2 engine steps to support job scheduling configuration.

Dashboard on Current Job-Scheduling Latency Benchmark:
[Kusto Dashboard Link](https://kusto.azure.com/dashboards/7e4753d7-f8be-45e0-b31c-267661a3e959?p-_startTime=1hours&p-_endTime=now#6350558f-03de-4762-bd75-5fee149cfcdf)

- As suggested by Robbie, set up cron job to monitor regression every 2 days, this will be monitored for at least 2 weeks before merging at job-scheduling.yml
- Added unit test at test_default_cli,py as suggest by @sumanthreddy29 

Pipeline 
https://dev.azure.com/akstelescope/telescope/_build/results?buildId=19154&view=results


vittoriasalim:

modules/python/clusterloader2/default/config/config.yaml
```diff
@@ -0,0 +1,59 @@
...
+          action: start
+          labelSelector: group=job-scheduling
+          threshold: 1h
+      - Identifier: WaitForFinishedJobs
+        Method: WaitForFinishedJobs
+        Params:
+          action: start
+          labelSelector: group=job-scheduling
+          threshold: 1h
+
```
wonderyl, line 31: Nitpicking: Remove 1 empty line?

modules/python/clusterloader2/utils.py
```diff
@@ -13,6 +13,7 @@
 NETWORK_METRIC_PREFIXES = ["APIResponsivenessPrometheus", "InClusterNetworkLatency", "NetworkProgrammingLatency"]
 PROM_QUERY_PREFIX = "GenericPrometheusQuery"
 RESOURCE_USAGE_SUMMARY_PREFIX = "ResourceUsageSummary"
+JOB_LIFECYCLE_LATENCY_PREFIX = {"JobLifecycleLatency_JobLifecycleLatency_":"JobLifecycleLatency_JobLifecycleLatency"}
```
wonderyl, line 16: IIUC, one of the JobLifecycleLatency is configured in config.yaml:19. How do we handle the case where the users specify a different Identifier of the measurement?

modules/python/csi/csi.py
```diff
@@ -22,7 +22,7 @@ def validate_node_count(node_label, node_count, operation_timeout_in_minutes):
             break
         print(f"Waiting for {node_count} nodes to be ready.")
         time.sleep(10)
-    if ready_node_count != node_count:
+    if ready_node_count < node_count:
```
wonderyl, line 25: Do you need to change this file? This won't be a problem when they don't use virtual nodes.

pipelines/perf-eval/CSI Benchmark/csi-attach-detach-300.yml
```diff
@@ -16,7 +16,6 @@ schedules:
 variables:
   SCENARIO_TYPE: perf-eval
   SCENARIO_NAME: storage-attach-detach-300
-
```
wonderyl, line 19: Do you need to change this?

pipelines/perf-eval/Controller/job-scheduling.yml
```diff
@@ -0,0 +1,33 @@
...
+          engine: clusterloader2
+          engine_input:
+              image: "ghcr.io/azure/clusterloader2:v20241022"
+          topology: kwok
+          matrix:
+              default:
+                  cpu_per_node: 1
+                  node_count: 20
+                  node_per_step: 5
+                  cilium_enabled: false
```
wonderyl, line 23: I wonder how can we avoid setting these parameters, we don't use these numbers at all.

wonderyl (commented):
Posting what I got so far, will continuing review a bit later.

wonderyl (commented):
Done my first round of review.

modules/python/clusterloader2/default/config/config.yaml
```diff
@@ -0,0 +1,59 @@
+name: job-scheduling
```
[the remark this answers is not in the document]
vittoriasalim: Somehow, the current configuration is unable to detect when CL2_CONFIG_DIR is placed outside the modules folder. Maybe we can create another folder inside modules/default to differentiate each test.
Will make a new pull request for this issue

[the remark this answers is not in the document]
wonderyl: Most likely the screnario folder wasn't copied to node. We can add that step.

modules/python/clusterloader2/utils.py
vittoriasalim: In the current configuration, we add new measurement to the prefix. These will then be processed by the get_measurement(file_path) function ([modules/python/clusterloader2/utils.py](https://github.com/Azure/telescope/pull/575/files/5a526730bfd98cafff3ca4fb2afdeeefcd0a2dd7#diff-e46631a8b31a54a4e3cb7da1eecbe362f53499f958d8dbd20fb8652e2b0b7d38)).

pipelines/perf-eval/Controller/job-scheduling.yml
vittoriasalim: Maybe we can reconfigure cli.py to accept optional parameters and process it differently. Will put that as a new task!

pipelines/perf-eval/CSI Benchmark/csi-attach-detach-300.yml
vittoriasalim: will revert it

vittoriasalim:
Update from previous version
Enabled a switch for virtual and actual node configuration in cli.py, allowing users to avoid entering non-needed parameters when operating in virtual mode.

modules/python/clusterloader2/default/config/config.yaml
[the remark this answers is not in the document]
vittoriasalim: Will make a new pull request for this.

scenarios/perf-eval/job-schedule/terraform-inputs/aws.tfvars
```diff
@@ -0,0 +1,92 @@
+scenario_type  = "perf-eval"
+scenario_name  = "job-schedule"
+deletion_delay = "20h"
```
sumanthreddy29, line 3: based on the pipeline runs the test completes in around 2hrs. Can you update this value to 3h?

scenarios/perf-eval/job-schedule/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,34 @@
+scenario_type  = "perf-eval"
+scenario_name  = "job-schedule"
+deletion_delay = "20h"
```
sumanthreddy29, line 3: same here. Update deletion time to 3h

modules/python/clusterloader2/default/cli.py
```diff
@@ -0,0 +1,766 @@
+import argparse
+import json
+import os
+import sys
+import time
+from datetime import datetime, timezone
+
+from kubernetes_client import KubernetesClient
+from utils import get_measurement, parse_xml_to_json, run_cl2_command
+
```
sumanthreddy29, line 10: what's the purpose of creating default folder under clusterloader2?

modules/python/clusterloader2/default/cli.py
```diff
@@ -0,0 +1,766 @@
...
+DEFAULT_PODS_PER_NODE = 40
+LOAD_PODS_PER_NODE = 20
+DEFAULT_NODES_PER_NAMESPACE = 100
+CPU_REQUEST_LIMIT_MILLI = 1
+DAEMONSETS_PER_NODE = {"aws": 2, "azure": 6, "aks": 6}
+CPU_CAPACITY = {"aws": 0.94, "azure": 0.87, "aks": 0.87}
+# TODO: Remove aks once CL2 update provider name to be azure
+
+
+def calculate_config(
```
sumanthreddy29, line 20: I would suggest to add python unit tests for the new logic that was implemented

modules/python/clusterloader2/default/cli.py
vittoriasalim: It was suggested to make one that is generic, so that other benchmark can reuse. As currently other benchmarks copy/paste their own version of the python code.

modules/python/clusterloader2/default/cli.py
```diff
@@ -0,0 +1,766 @@
...
+            file.write("CL2_PROMETHEUS_SCRAPE_CILIUM_AGENT: true\n")
+            file.write("CL2_PROMETHEUS_SCRAPE_CILIUM_AGENT_INTERVAL: 30s\n")
+
+        if service_test:
+            file.write("CL2_SERVICE_TEST: true\n")
+        else:
+            file.write("CL2_SERVICE_TEST: false\n")
+
+        if cnp_test:
+            file.write("CL2_CNP_TEST: true\n")
```
wonderyl, line 155: Is this supported by CL2?

vittoriasalim: Yes, we're not using that in job scheduling, so the parameters aren't included in the template ([pipelines/perf-eval/Controller/job-scheduling.yml](https://github.com/Azure/telescope/pull/575/files#diff-36befacf6bced05d1871d3ebc34f4b554477980c56ca31742b9a141ee7ff9bb3)).

Since cli.py is intended to be generic, I'm including this to support the CNP test type when users need to run it -- just like the initial configuration. If CNP is not included, it will default to false.

modules/python/clusterloader2/default/cli.py
```diff
@@ -0,0 +1,766 @@
...
+):
+    with open(override_file, "w", encoding="utf-8") as file:
+        file.write(f"PROVIDER: {provider}\n")
+        file.write(f"CL2_OPERATION_TIMEOUT: {operation_timeout}\n")
+        file.write("CL2_PROMETHEUS_TOLERATE_MASTER: true\n")
+        file.write("CL2_PROMETHEUS_MEMORY_LIMIT_FACTOR: 30.0\n")
+        file.write("CL2_PROMETHEUS_MEMORY_SCALE_FACTOR: 30.0\n")
+        file.write('CL2_PROMETHEUS_NODE_SELECTOR: "prometheus: \\"true\\""\n')
+        file.write("CL2_POD_STARTUP_LATENCY_THRESHOLD: 3m\n")
+        file.write(f"CL2_KWOK_NODES: {kwok_nodes}\n")
```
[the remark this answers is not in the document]
vittoriasalim: It is used to pass parameters to the CL2 config, which then determines how many virtual nodes are needed to run in the configuration ([config.yaml](https://github.com/Azure/telescope/pull/575/files#diff-d611201cc8ef772003f543e2144fb20eabc2b2dabd438e72ce297718eb817dd5)).

This also applies when Prometheus is enabled ([load-config.yaml](https://github.com/Azure/telescope/blob/main/modules/python/clusterloader2/slo/config/load-config.yaml)).

modules/python/clusterloader2/utils.py
wonderyl: also the name is "JOB_LIFECYCLE_LATENCY_PREFIX ", which makes it hard to reuse by other benchmarks.

pipelines/perf-eval/Controller/job-scheduling.yml
```diff
@@ -0,0 +1,60 @@
...
+      regions:
+      - us-east-2
+      engine: clusterloader2
+      engine_input:
+        image: "ghcr.io/azure/clusterloader2:v20241022"
+      topology: kwok
+      matrix:
+        default:
+          kwok_nodes: 2000
+          qps: 60
```
wonderyl, line 51: why the qps of aws is different than azure?

pipelines/perf-eval/Controller/job-scheduling.yml
```diff
@@ -0,0 +1,60 @@
...
+      engine: clusterloader2
+      engine_input:
+        image: "ghcr.io/azure/clusterloader2:v20241022"
+      topology: kwok
+      matrix:
+        default:
+          kwok_nodes: 2000
+          qps: 10
+          job_count: 20000
+          cpu_per_node: 1
```
wonderyl, line 26: Do you still need cpu_per_node?

modules/python/clusterloader2/default/cli.py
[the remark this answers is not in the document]
wonderyl: Is CL2_KWOK_NODES used in config.yaml? 
```
name: job-scheduling
{{$job_count := DefaultParam .CL2_JOB_COUNT 20000}}
{{$qps := DefaultParam .CL2_QPS 200}}
```

pipelines/perf-eval/Controller/job-scheduling.yml
vittoriasalim: Right, I missed it. removed

steps/engine/clusterloader2/default/collect.yml
```diff
@@ -0,0 +1,30 @@
...
+  parameters:
+    region: ${{ parameters.region }}
+- script: |
+    echo "Collecting results -- "
+    ls $CL2_REPORT_DIR
+    set -eo pipefail
+
+    PYTHONPATH=$PYTHONPATH:$(pwd) python3 $PYTHON_SCRIPT_FILE collect virtual\
+      $KWOK_NODES $QPS $JOB_COUNT \
+      $CL2_REPORT_DIR "$CLOUD_INFO" $RUN_ID $RUN_URL $SERVICE_TEST ${CNP_TEST:-False} ${CCNP_TEST:-False} ${NUM_CNPS:-0} ${NUM_CCNPS:-0} ${DUALSTACK:-False} $TEST_RESULTS_FILE \
```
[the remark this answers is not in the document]
vittoriasalim: It has been fixed.

I’ve also added support for running both virtual and actual node configurations:

python3 cli.py configure virtual [...] — runs in virtual node
python3 cli.py configure actual [...] — runs in actual node

# if no mode is specified
python3 cli.py configure [...] — defaults to actual node

modules/python/clusterloader2/default/cli.py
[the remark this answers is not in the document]
vittoriasalim: Right — parameters from the [job-scheduling pipeline](https://github.com/Azure/telescope/blob/vitto/kwok-cl2/pipelines/perf-eval/Controller/job-scheduling.yml) are correctly passed into the CL2 config.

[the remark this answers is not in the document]
vittoriasalim: My bad,
Right now, we’re only using` job_count` and `qps` in `config.yaml`.

I assumed, kwok_nodes will be needed for scaling or to set up different CL2 configurations — since cli.py will be generic.

For prometheus, it will only be passed only when prometheus is enabled -- based on initial configurations
```
run_cl2_command(
        kubeconfig,
        cl2_image,
        cl2_config_dir,
        cl2_report_dir,
        provider,
        cl2_config_file=cl2_config_file,
        overrides=True,
        enable_prometheus=False,
    )
```

modules/python/clusterloader2/default/cli.py
```diff
@@ -0,0 +1,766 @@
...
+        execute_clusterloader2(
+            args.cl2_image,
+            args.cl2_config_dir,
+            args.cl2_report_dir,
+            args.cl2_config_file,
+            args.kubeconfig,
+            args.provider,
+        )
+    elif args.command == "collect":
+        if args.mode == "virtual":
```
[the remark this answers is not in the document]
vittoriasalim: Right — virtual mode will simulate nodes, while non-virtual mode will run on actual nodes.

I think that will be great; I’ll give it a try. Currently, we have separate functions for virtual and actual nodes since they require different sets of arguments and templates. I suppose it would be a bit difficult to handle both with a single function, as some arguments used in virtual mode don’t exist for actual nodes. I will follow up on this.

pipelines/perf-eval/Controller/job-scheduling.yml
vittoriasalim: Thanks, fixed!

modules/python/clusterloader2/default/cli.py
sumanthreddy29: Can we create a clusterloaderbase class and add the generic methods into it and use it. If anyone wants to customize then can extend on top of this base class?

steps/engine/clusterloader2/default/collect.yml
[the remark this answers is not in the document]
wonderyl: The way you fixed it is to provide default value on the client side, it's less preferred than providing on the server side, because the server knows best what default value it should be. What you can do it to use named parameters e.g. `--cnp-test` and provide default value to these parameters in the cli.py script. Then on the client side, i.e. yaml file, you can avoid setting it when it's not needed.

modules/python/clusterloader2/default/cli.py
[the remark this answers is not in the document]
wonderyl: We set up kwok nodes externally. CL2 is agnostic to Kwok. Thus virtual is not relevant to CL2. you can add a parameter to not set up nodes though.

modules/python/clusterloader2/default/cli.py
wonderyl: Ideally someone who just write a pipeline doesn't need to rewrite any python code for CL2, right? CL2 can be configured using configs already. It's really not necessary to create a new python class every time.

steps/engine/clusterloader2/default/collect.yml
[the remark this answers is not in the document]
vittoriasalim: Actually, I provided the default value on both the client and server sides. I can remove the one from client side, since it won't have any effect.

`parser_collect_virtual.add_argument("service_test", type=eval, choices=[True, False], default=False, help="Whether service test is running. Must be either True or False")
    parser_collect_virtual.add_argument("cnp_test", type=eval, choices=[True, False], nargs="?", default=False, help="Whether cnp test is running. Must be either True or False")
    parser_collect_virtual.add_argument("ccnp_test", type=eval, choices=[True, False], nargs="?", default=False, help="Whether ccnp test is running. Must be either True or False")
    parser_collect_virtual.add_argument("num_cnps", type=int, nargs="?", default=0, help="Number of cnps")
    parser_collect_virtual.add_argument("num_ccnps", type=int, nargs="?", default=0, help="Number of ccnps")
    parser_collect_virtual.add_argument("dualstack", type=eval, choices=[True, False], nargs="?", default=False, help="Whether cluster is dualstack. Must be either True or False")`

modules/python/clusterloader2/default/cli.py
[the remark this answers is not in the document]
vittoriasalim: Got it.

vittoriasalim:
@wonderyl , Can I get a review, I have updated this PR, 
- Modified cli.py to accept an optional parameter using the -- syntax.
- Moved the job scheduler cl2 configuration directory to the scenarios folder.
- Updated start-kwok to include a status condition, to ensure fake nodes have a Ready condition for node validation.
- Updated cli.py to accept both pod and job workload types (previously, configure_clusterloader2 and collect_cl2 was heavily configured just for pod deployment).
- Extended the deletion timeout to 3 hours, as suggested by @sumanthreddy29.
- Added basic unit tests to test_cli.py (more edge cases will be added later), as suggested by @sumanthreddy29.
- Set up a cron job as suggested by Robbie to monitor job scheduler regression every 2 days. --- This will be monitored for at least 2 weeks before merging. ---
- Included black formatter and isort.
- Updated working directory to support new changes from main (kubernetes_clients has been moved to separate folder)

modules/python/tests/test_cli.py
```diff
@@ -0,0 +1,168 @@
+import json
```
wonderyl, line 1: Would it better to name it `test_cl2_default_cli.py` instead? the test folder is shared among all python modules.

scenarios/perf-eval/job-scheduler/config/config.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+namespace:
+  number: 1
+  prefix: job-scheduling
+  deleteStaleNamespaces: true
+  deleteAutomanagedNamespaces: true
+  enableExistingNamespaces: false
+
+tuningSets:
+- name: Uniform1qps
```
wonderyl, line 14: Should change the name? it's not 1 qps anymore

pipelines/perf-eval/Controller/job-scheduling.yml
```diff
@@ -1,32 +1,66 @@
...
+    displayName: "1:30 AM every 2 days"
+    branches:
+      include:
+        - main
+        - vitto/kwok-cl2 # to be removed after the PR is merged
+    always: true
 variables:
   SCENARIO_TYPE: perf-eval
-  SCENARIO_NAME: job-scheduling
+  SCENARIO_NAME: job-scheduler
```
wonderyl, line 12: In k8s, scheduler refer to the pod scheduler. We are testing the capability of "job scheduling"

modules/python/tests/test_cli.py
vittoriasalim: Will make an update on that!

modules/python/clusterloader2/utils.py
vittoriasalim: Will make a new PR for this, since the change will affect other pipelines

modules/python/tests/test_cli.py
```diff
@@ -0,0 +1,168 @@
...
+            "job_throughput": 1000,
+        }
+
+        # Act
+        configure_clusterloader2(**mock_data)
+
+        # Assert
+        # Verify that the file was opened for writing
+        mock_open_file.assert_any_call(
+            mock_data["cl2_override_file"], "w", encoding="utf-8"
```
wonderyl, line 52: It's usually a bad sign to test behaviors, as that's implementation details. A test should test the function contract. In this case,  configure() generates a config file based on the input, so you should test the content of the config file is as expected.

modules/python/tests/test_cli.py
```diff
@@ -0,0 +1,168 @@
...
+        node_count = 2  # Expected number of ready nodes
+        operation_timeout_in_minutes = 5
+
+        # Act
+        validate_clusterloader2(node_count, operation_timeout_in_minutes)
+
+        # Assert: verify if get_ready_nodes was called
+        mock_get_ready_nodes.assert_called_once()
+        mock_kubernetes_client_init.assert_called_once()
+        self.assertEqual(mock_get_ready_nodes.return_value, ["node1", "node2"])
```
wonderyl, line 85: This assertion is useless, since you set the return value yourself.

modules/python/tests/test_cli.py
```diff
@@ -0,0 +1,168 @@
...
+    ):
+        # Arrange
+        node_count = 2  # Expected number of ready nodes
+        operation_timeout_in_minutes = 5
+
+        # Act
+        validate_clusterloader2(node_count, operation_timeout_in_minutes)
+
+        # Assert: verify if get_ready_nodes was called
+        mock_get_ready_nodes.assert_called_once()
```
wonderyl, line 83: It's the same problem as the last one. The contract of the validate_clusterloader2 function is, it will throw an exception if the node count is less than expected after timeout runs out. Then you should test when an exception is thrown and when it doesn't.

vittoriasalim:
@wonderyl 
I have made the suggested corrections:
- Renamed test_cli.py to test_default_cli.py.
- Redesigned test_default_cli.py based on the suggestion to check the behavior of the function and avoid using useless assertions.
- Renamed uniform1qps to uniform{$qps}qps, since it is no longer 1qps.
- Renamed job scheduler to job scheduling.

modules/python/tests/test_default_cli.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+class TestCollectClusterLoader2(unittest.TestCase):
+    def test_collect_clusterloader2(self):
+        # Create a temporary directory for the report
+        cl2_report_dir = os.path.join(
+            os.path.dirname(__file__), "mock_data", "default", "report"
+        )
+        # Create a temporary file for result output
+        result_file = tempfile.mktemp()
```
wonderyl, line 119: mktemp() is deprecated https://docs.python.org/3/library/tempfile.html#deprecated-functions-and-variables

modules/python/tests/test_default_cli.py
```diff
@@ -0,0 +1,187 @@
...
+    collect_clusterloader2,
+    configure_clusterloader2,
+    validate_clusterloader2,
+)
+
+
+class TestConfigureClusterLoader2(unittest.TestCase):
+    def test_configure_clusterloader2(self):
+        # Create a temporary file for the override file
+        with tempfile.NamedTemporaryFile(
```
wonderyl, line 17: Would `tempfile.mkstemp()` be a better option here? since you need to delete it yourself.

vittoriasalim:
@wonderyl  Converted mktemp() to mkstemp()

wonderyl (approved):

.yamllint
```diff
@@ -20,3 +19,4 @@ rules:
 ignore: |
   modules/python/clusterloader2/**/*.yaml
   modules/python/clusterloader2/**/*.yml
+  scenarios/perf-eval/job-scheduling/**/*.yaml
```
alyssa1303, line 22: Those files should be moved to clusterloader2 instead of adding ignore here

modules/python/clusterloader2/default/cli.py
alyssa1303: This is not necessarily true because different test requires only a specific set of arguments and configs, so combining this all together like this make it hard for new person to come and modify code if they only want to use a subset of arguments and configs. It's better to have base class like Sumanth suggests so other test can extend, which is not something this script is doing. So for this test, I'd suggest just add whatever argument you need for your test, and in another PR, create a base class for ClusterLoader2 and refactor

scenarios/perf-eval/job-scheduling/config/config.yaml
```diff
@@ -0,0 +1,62 @@
+name: job-scheduling
```
alyssa1303, line 1: this should be moved to clusterloader2 folder for re-use

scenarios/perf-eval/job-scheduling/config/job_template.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: batch/v1
```
alyssa1303, line 1: this should be moved to clusterloader2 folder for re-use

steps/engine/clusterloader2/default/cleanup.yml
```diff
@@ -0,0 +1,6 @@
+steps:
```
alyssa1303, line 1: why is this file needed? Are you using karpenter or create new nodepool and nodeclaim?

steps/topology/kwok/start-kwok.yml
```diff
@@ -0,0 +1,69 @@
+parameters:
+  - name: kwok_repo
+    type: string
+    default: "kubernetes-sigs/kwok"
+  - name: kwok_release
+    type: string
+    default: "v0.6.1"
+
+steps:
+  - script: |
```
alyssa1303, line 10: can this be done in python script?

pipelines/perf-eval/Controller/job-scheduling.yml
```diff
@@ -0,0 +1,66 @@
...
+    dependsOn: []
+    jobs:
+      - template: /jobs/competitive-test.yml
+        parameters:
+          cloud: azure
+          regions:
+            - eastus2
+          engine: clusterloader2
+          engine_input:
+            image: "ghcr.io/azure/clusterloader2:v20241022"
```
sumanthreddy29, line 25: We have recently merged upstream changes into perf-tests repo and generated a new image for cl2. Can you use the latest image and run your tests on the latest image?
```suggestion
            image: "ghcr.io/azure/clusterloader2:v20250423"
```

pipelines/perf-eval/Controller/job-scheduling.yml
```diff
@@ -0,0 +1,66 @@
+trigger: none
+schedules:
+  - cron: "30 1 */2 * *" # Every 2 days at 1:30 AM
+    displayName: "1:30 AM every 2 days"
+    branches:
+      include:
+        - main
+        - vitto/kwok-cl2 # to be removed after the PR is merged
```
sumanthreddy29, line 8: private branch scheduled should be removed before the merge. If not, you have to create new PR again to remove this branch.

sumanthreddy29: I would suggest leaving this comment unresolved till you resolve all other comments and remove this change after you get an approval for this PR.

pipelines/perf-eval/Controller/job-scheduling.yml
```diff
@@ -0,0 +1,66 @@
...
+    dependsOn: []
+    jobs:
+      - template: /jobs/competitive-test.yml
+        parameters:
+          cloud: aws
+          regions:
+            - us-east-2
+          engine: clusterloader2
+          engine_input:
+            image: "ghcr.io/azure/clusterloader2:v20241022"
```
sumanthreddy29, line 51:
```suggestion
            image: "ghcr.io/azure/clusterloader2:v20250423"
```
Test your changes with newer image.

steps/topology/kwok/collect-clusterloader2.yml
```diff
@@ -0,0 +1,18 @@
+parameters:
+- name: cloud
+  type: string
+  default: ''
+- name: engine_input
+  type: object
+  default: {}
+- name: regions
+  type: object
+
```
sumanthreddy29, line 10: extra space

steps/topology/kwok/start-kwok.yml
vittoriasalim: Python script inside cl2?

.yamllint
wonderyl: Here's the thought process. Previously cl2 configs are under modules folder. The problem with that is, to write a new config for a new pipeline, we need to copy/past a whole folder of code. This not only include the config files, but also the python files as well. This is very poor in terms of usability. So we propose a new solution, moving config files under scenarios, and created a default python module, that can be reused. Then creating a new pipeline doesn't require to copy the python files anymore.

modules/python/clusterloader2/default/cli.py
wonderyl: You can set the arguments in execute.yaml and we moved the configs under scenarios folder. This would avoid changing any python code, no?

PR #576, opened 2025-03-25
Collect 50/70/90/99/100 Percentile Node Latency for Cluster Autoscaler & NAP tests

sumanthreddy29:
1. Add steps to collect 50/70/90/99/100 Percentile Node Latency for Cluster Autoscaler & NAP tests


modules/python/clusterloader2/autoscale/autoscale.py
```diff
@@ -113,7 +113,7 @@ def collect_clusterloader2(
     run_id,
     run_url,
     result_file
-):
+): # pylint: disable=too-many-branches
```
alyssa1303, line 116: can we add helper function to reduce the if else branch?

modules/python/clusterloader2/autoscale/autoscale.py
```diff
@@ -141,13 +141,37 @@ def collect_clusterloader2(
...
+                summary[index]["up"]["wait_for_70Perc_nodes_seconds"] = -1 if failure else testcase["time"]
+                summary[index]["up"]["failures"] += 1 if failure else 0
+            elif "WaitForNodesUpPerc90" in name:
+                summary[index]["up"]["wait_for_90Perc_nodes_seconds"] = -1 if failure else testcase["time"]
+                summary[index]["up"]["failures"] += 1 if failure else 0
+            elif "WaitForNodesUpPerc99" in name:
+                summary[index]["up"]["wait_for_99Perc_nodes_seconds"] = -1 if failure else testcase["time"]
+                summary[index]["up"]["failures"] += 1 if failure else 0
+            elif "WaitForNodesUpPerc100" in name:
                 summary[index]["up"]["wait_for_nodes_seconds"] = -1 if failure else testcase["time"]
```
alyssa1303, line 157: It seems we might need to update the json format due to new columns. Then maybe we should change `wait_for_nodes_seconds` to `wait_for_100Perc_nodes_seconds` for consistency?

modules/python/clusterloader2/autoscale/config/config.yaml
```diff
@@ -52,7 +56,43 @@ steps:
       labelSelector: {{$podLabelSelector}}
       timeout: {{$scaleUpTimeout}}
       refreshInterval: {{$refreshInterval}}
-  - Identifier: WaitForNodesUp {{$i}}
+  - Identifier: WaitForNodesUpPerc50 {{$i}}
```
alyssa1303, line 59: I wonder if we can do loop here with a variable to adjust and calculate percentile instead of copy and paste same methods?

modules/python/clusterloader2/autoscale/autoscale.py
sumanthreddy29: updated

modules/python/clusterloader2/autoscale/config/config.yaml
sumanthreddy29: Added loop

alyssa1303 (approved):

PR #590, opened 2025-04-07
Expand matrix to cover 250 pods on 32 cpu node for Azure

alyssa1303:


pipelines/perf-eval/CRI Benchmark/cri-resource-consume.yml
```diff
@@ -151,6 +151,16 @@ stages:
               k8s_machine_type: Standard_D16ds_v4
               k8s_os_disk_type: Ephemeral
               scrape_kubelets: True
+            n10-p2500-memory-1-31:
```
sumanthreddy29, line 154: Do we need to add the same for 1.30 k8s version?

sumanthreddy29 (approved):

pipelines/perf-eval/CRI Benchmark/cri-resource-consume.yml
alyssa1303: We agree to expand in another PR if needed

PR #591, opened 2025-04-07
Extract resource creation & deletion metrics from terraform operations for each resource

sumanthreddy29:
1. Save terraform apply/destroy command logs to log file
2. Process the logs and extract resource creation/deletion time for each resource.
3. Push the results to storage account under system container.


modules/python/terraform/extract_terraform_operation_metadata.py
```diff
@@ -0,0 +1,87 @@
...
+import sys
+import os
+import datetime
+
+# Regex to extract: full module path and time
+PATTERN = re.compile(r"(module\.[^:]+): (?:Creation|Destruction) complete after (\d+m\d+s|\d+s)")
+
+def time_to_seconds(time_str):
+    try:
+        if 'm' in time_str:
```
alyssa1303, line 12: Can we check format for when creation or deletion took more than an hour (although unlikely but there can be stuck resources) so we can parse such case too?

modules/python/tests/test_extract_terraform_operation_metadata.py
```diff
@@ -0,0 +1,158 @@
...
+from unittest.mock import patch, mock_open
+
+from terraform.extract_terraform_operation_metadata import (
+  time_to_seconds,
+  parse_module_path,
+  process_terraform_logs,
+)
+
+class TestExtractTerraformOperationMetadata(unittest.TestCase):
+    def test_time_to_seconds_with_minutes_and_seconds(self):
```
alyssa1303, line 12: should add a case for when it takes a long time (>1 hour)

steps/collect-terraform-operation-metadata.yml
```diff
@@ -0,0 +1,29 @@
...
+    echo "##vso[task.setvariable variable=TERRAFORM_OPERATION_METADATA_FILE]$TEST_RESULTS_DIR/terraform_operation_metadata.json"
+  displayName: "Collect Terraform Operation Metadata"
+  condition: ne(variables['SKIP_RESOURCE_MANAGEMENT'], 'true')
+  workingDirectory: modules/python/terraform
+  env:
+    PYTHON_SCRIPT_FILE: $(Pipeline.Workspace)/s/modules/python/terraform/extract_terraform_operation_metadata.py
+
+- ${{ if eq(variables['Build.SourceBranchName'], 'main') }}:
+  - template: /steps/cloud/azure/upload-storage-account.yml
+    parameters:
```
alyssa1303, line 22: should also add the condition for when skip resource management is not equal to true. Otherwise, we won't have file to upload

steps/terraform/run-command.yml
```diff
@@ -91,3 +91,8 @@ steps:
       ARM_USE_MSI: true
       ARM_TENANT_ID: $(AZURE_MI_TENANT_ID)
       ARM_CLIENT_ID: $(AZURE_MI_CLIENT_ID)
+- ${{ if eq(parameters.command, 'destroy') }}:
```
alyssa1303, line 94: why do we need this step? Doesn't the above take into account destroy already?

modules/python/tests/test_extract_terraform_operation_metadata.py
sumanthreddy29: updated

steps/collect-terraform-operation-metadata.yml
sumanthreddy29: I added condition at the entry level in the run-command, so that all the steps below  them will skip

steps/terraform/run-command.yml
sumanthreddy29: We want to collect data after we run apply/destroy command, so that it doesn't run for each command.

alyssa1303 (approved):

PR #598, opened 2025-04-17
Add terraform and SSH  to pypeline

vittoriasalim:
- Add terraform 
- Add SSH


pypeline/benchmark.py
```diff
@@ -19,40 +19,49 @@ def validate(self) -> list[Step]:
...
+        pass
 
+    @abstractmethod
+    def get_region(self) -> str:
+        pass
+
+    @abstractmethod
+    def get_credential_type(self) -> str:
+        pass
 @dataclass
```
wonderyl, line 43: Is this file formatted by a tool? It looks to me the format is not right.

pypeline/benchmark.py
```diff
@@ -19,40 +19,49 @@ def validate(self) -> list[Step]:
...
     def run(self) -> list[Step]:
         pass
 
-
 class Cloud(ABC):
     @abstractmethod
     def login(self) -> list[Step]:
         pass
 
+    @abstractmethod
```
wonderyl, line 32: I wonder if you can move these fields as data fields of the Cloud class, then you don't need expose these getters anymore. I'm trying to avoid these boiler plates by using data class directly.

pypeline/resource/terraform/terraform.py
```diff
@@ -0,0 +1,184 @@
...
+        """.strip(
+            "\n"
+        )
+    ),
+    condition=f"and(eq('{cloud}', 'azure'), ne(variables['SKIP_RESOURCE_MANAGEMENT'], 'true'))",
+    env={"region": regions},
+)
+
+
+class Terraform(Resource):
```
wonderyl, line 130: Could you make this a dataclass to avoid the boilerplate of init?

pypeline/resource/terraform/run_command.py
```diff
@@ -0,0 +1,89 @@
...
+        "REGIONS": regions,
+        "CLOUD": cloud,
+        "ARM_SUBSCRIPTION_ID": "$(AZURE_SUBSCRIPTION_ID)",
+        **(
+            {
+                "ARM_USE_MSI": "true",
+                "ARM_TENANT_ID": "$(AZURE_MI_TENANT_ID)",
+                "ARM_CLIENT_ID": "$(AZURE_MI_CLIENT_ID)",
+            }
+            if credential_type == "managed_identity"
```
wonderyl, line 85: Could you make credential type a Enum instead to avoid typo?

pypeline/cloud/azure.py
```diff
@@ -86,3 +86,14 @@ def login(self) -> list[Step]:
...
                 ),
             ]
+
+    def get_cloud_type(self) -> str:
+        return "azure"
+
+    def get_region(self) -> str:
+        return self.region
+
+    def get_credential_type(self) -> str:
```
wonderyl, line 96: Is this a property of the Cloud or a choice of the pipeline?

pypeline/pipeline.py
```diff
@@ -16,7 +16,11 @@ class Step:
 class Script(Step):
     display_name: str = field(metadata={"yaml": "displayName"})
     script: str
-    env: Optional[dict[str, str]] = None
+    condition: Optional[str] = field(metadata={"yaml": "condition"}, default=None)
```
wonderyl, line 19: You only need to set metadata yaml when the output name is different than the variable name. E.g. the variable is display_name while the output yaml name is displayName.

pypeline/resource/terraform/input_variables.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    ).strip(),
+    env={
+        "CLOUD": cloud,
+        "REGIONS": regions,
+        "RUN_ID": "$(RUN_ID)",
+        "INPUT_VARIABLES": input_variables,
+        "DEBUG": "$(System.Debug)",
+    },
+    condition="ne(variables['SKIP_RESOURCE_MANAGEMENT'], 'true')",
```
wonderyl, line 52: Could we remove the env variable and make it an input to a resource? We should eliminate all env variables. It's really hard to track env var usage, and it cut though call stack without any trace.

pypeline/resource/terraform/input_variables.py
```diff
@@ -0,0 +1,185 @@
+from textwrap import dedent
+from pipeline import Script
+
+set_input_variables_aws = lambda cloud, regions, input_variables={}: Script(
```
wonderyl, line 4: Could you make it a regular function? lambdas are usually used for short functions.

wonderyl, line 4: Could you use typing please?

wonderyl, line 4: Does it make sense to write aws, azure and gcp terraforms in 3 different classes?

pypeline/resource/terraform/input_variables.py
```diff
@@ -0,0 +1,185 @@
...
+          echo "Input variables: $input_variables_str"
+          regional_config=$(echo "$regional_config" | jq --arg region "$REGION" --arg input_variable "$input_variables_str" \
+            '.[$region].TERRAFORM_INPUT_VARIABLES += $input_variable')
+          echo "Regional config: $regional_config"
+          INPUT_VARIABLES=""
+        done
+        regional_config_str=$(echo $regional_config | jq -c .)
+        echo "Final regional config: $regional_config_str"
+        echo "##vso[task.setvariable variable=TERRAFORM_REGIONAL_CONFIG]$regional_config_str"
+    """
```
wonderyl, line 43: Should you align the """?

wonderyl, line 43: Should you align the """?

pypeline/resource/terraform/input_variables.py
```diff
@@ -0,0 +1,185 @@
...
+    ).strip(),
+    env={
+        "CLOUD": cloud,
+        "REGIONS": regions,
+        "RUN_ID": "$(RUN_ID)",
+        "INPUT_VARIABLES": input_variables,
+        "DEBUG": "$(System.Debug)",
+        "SKU_TIER": "$(SKU_TIER)",
```
wonderyl, line 123: Ditto for these env vars. These should be input variables to a Resource.

pypeline/resource/terraform/terraform.py
wonderyl, line 130: Technically Terraform is not a resource, AKS is a resource. We use Terraform to set up an AKS cluster. I think Terraform could be a good base class that other resources inherit and reuse its implementations.

pypeline/pipelines/controller/job_scheduling.py
```diff
@@ -5,22 +5,27 @@
...
+    cloud_az_eastus2 = Azure(region="eastus2")
+    azure_east_us2 = Layout(
+        display_name="Job Scheduling",
+        setup=Setup(run_id=os.getenv("RUN_ID")),
+        cloud=cloud_az_eastus2,
+        resources=[
+            Python3(),
+            Terraform(
+                cloud_obj=cloud_az_eastus2
+            ),  # Dynamically retrieve cloud from layout
```
wonderyl, line 21: What does this comment mean?, are you talking about line 20?

wonderyl, line 21: What does this comment mean?, are you talking about line 20?

pypeline/resource/terraform/terraform.py
```diff
@@ -0,0 +1,184 @@
...
+    cloud_obj : Cloud = None
```
wonderyl, line 131: Should you just name it cloud? obj serves no meaning.

vittoriasalim: Right!

pypeline/resource/terraform/run_command.py
```diff
@@ -0,0 +1,89 @@
...
+    display_name=f"Run Terraform {command.capitalize()} Command",
+    script=dedent(
+        f"""
+        set -e
+
+        # Navigate to the Terraform working directory
+        cd $TERRAFORM_WORKING_DIRECTORY
+
+        # Handle apply or destroy commands
+        if [[ "{command}" == "apply" || "{command}" == "destroy" ]]; then
```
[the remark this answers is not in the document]
vittoriasalim: Right. Better make it out of the script.

pypeline/resource/terraform/input_variables.py
vittoriasalim: Right, Noted

pypeline/cloud/azure.py
vittoriasalim: property of cloud, will move it to cloud class to remove getters

pypeline/resource/terraform/input_variables.py
vittoriasalim: When I checked the Terraform Resource, the only place where we currently need separate commands for AWS, Azure, and GCP is in the set_input_variables function. 

I suppose for now, it’s best to keep them within Terraform for maintainability and to avoid high coupling—until it’s extended further. What do you think?

image

pypeline/cloud/azure.py
wonderyl: What I meant was, how you sign in is not a property of a Cloud class, it's a user's choice. You can pass it in as an option of login or other functions that use it.

pypeline/resource/terraform/input_variables.py
wonderyl: SGTM, let's how they are used in an actual resource to decide.

vittoriasalim:
@wonderyl 

Addition :
- Added SSH resource
- Set up a resource factory to avoid reinitializing shared properties for every resource.
- Moved all components to components.py to avoid circular imports and to keep all components in one place.
Corrections
- Removed environment variables and made them inputs to resources.
- Refactored Terraform into its own class, refactor run_command and input_variables
- Ensured proper typing and converted lambda expressions into functions.
- Used an Enum for credential types.
- Moved logic outside the script for better readability.
- Moved properties into data fields within the cloud class to avoid exposing getters and setters.

.gitignore
```diff
@@ -48,6 +48,8 @@ terraform.*
 .terraform*
 darwin-amd64/
 linux-amd64/
+# Exception: Do not ignore terraform.py
+!terraform.py
```
wonderyl, line 52: It's better to write line 48 more specific and avoid this exception.
E.g.
```
.terraform/
```
This will not ignore terraform.py

pypeline/components.py
```diff
@@ -0,0 +1,58 @@
...
+# Keep components here to avoid circular imports
+
+
+class CredentialType(Enum):
+    MANAGED_IDENTITY = "managed_identity"
+    SERVICE_CONNECTION = "service_connection"
+
+@dataclass
+class Resource(ABC):
+    cloud: str
```
wonderyl, line 16: A resource can be a python binary, that's not related to Cloud at all.

pypeline/components.py
```diff
@@ -0,0 +1,58 @@
...
+    regions: list[str]
```
wonderyl, line 17: Ditto

pypeline/components.py
```diff
@@ -0,0 +1,58 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from enum import Enum
+
+from pipeline import Step
+
+# Keep components here to avoid circular imports
```
wonderyl, line 7: Could you elaborate more about this?

pypeline/components.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @abstractmethod
+    def tear_down(self) -> list[Step]:
+        pass
+
+
+# Factory class to create resources:
+# To avoid repeatedly define shared properties for every resource. usage: @ job_scheduling.py
+@dataclass
+class ResourceFactory:
```
wonderyl, line 35: This is not a factory pattern. https://en.wikipedia.org/wiki/Factory_method_pattern
As mentioned before, not all resource is associate with a cloud or a region.

pypeline/resource/setup.py
```diff
@@ -1,96 +1,115 @@
...
-            "\n"
-        )
-    ),
-    env={"TEST_MODULES_DIR": test_modules_dir},
-)
+
+
+def set_run_id(run_id: str) -> Script:
+    if not run_id:
+        run_id = "$(Build.BuildId)-$(System.JobId)"
```
wonderyl, line 10: $(Build.BuildId) and $(System.JobId) azure pipeline's system env vars. We should not extract them out. In addition, it doesn't work, we don't define $(Build.BuildId) in Python.

pypeline/resource/setup.py
```diff
@@ -1,96 +1,115 @@
...
+            """
+        ).strip(),
+        condition="eq(variables['SKIP_RESOURCE_MANAGEMENT'], 'true')",
+    )
 
 
 @dataclass
 class Setup(Resource):
     run_id: str
+    engine: str
```
wonderyl, line 96: Where is engine used?

pypeline/resource/setup.py
```diff
@@ -1,96 +1,115 @@
...
     test_module_dir: str = ""
+    engine_input: dict = field(default_factory=dict)
```
wonderyl, line 98: Where is engine_input used?

pypeline/resource/setup.py
```diff
@@ -1,96 +1,115 @@
...
+
+
+def set_script_module_directory(test_modules_dir: str) -> Script:
+    if not test_modules_dir:
+        test_modules_dir = "modules/bash"
+
+    script_modules_directory = f"$(Pipeline.Workspace)/s/{test_modules_dir}"
+
+    return Script(
+        display_name="Set Script Module Directory",
```
wonderyl, line 67: We still need to get `$(Pipeline.Workspace)` from the Azure pipeline env to get the full path of test_modules_directory. Then it sets a variable TEST_MODULES_DIR to record this value, which will be used in later steps. I don't think we can remove it.

wonderyl: You need to distinguish 2 types of variables.
1 is used set in the env variable. e.g. `env={"TEST_MODULES_DIR": test_modules_dir},` we set this, so we can eliminate them. 
2 is `##vso[task.setvariable variable=TEST_MODULES_DIR]` these are azure pipeline variables, we can't remove them.

pypeline/resource/setup.py
```diff
@@ -1,96 +1,115 @@
...
-            set_test_results_directory,
+            set_run_url_and_code_url(),
+            set_test_results_directory(),
             set_script_module_directory(self.test_module_dir),
         ]
 
     def validate(self) -> list[Step]:
+        return [validate_owner_info()]
+
+    def execute_tests(self) -> list[Step]:
```
wonderyl, line 111: Where is this method used?

pypeline/cloud/azure.py
```diff
@@ -1,99 +1,80 @@
...
-    MANAGED_IDENTITY = "managed_identity"
-    SERVICE_CONNECTION = "service_connection"
-
-
 @dataclass
 class Azure(Cloud):
-    region: str = "eastus"
+    cloud: str = "azure"
+    regions: list[str] = field(default_factory=lambda: ["eastus2"])
     subscription: str = os.getenv("AZURE_SUBSCRIPTION_ID")
```
wonderyl, line 15: I wonder why we need to get these vars from env vars, rather than pass them in when creating an Azure object?

vittoriasalim: I checked some variables like AZURE_SUBSCRIPTION_ID, SERVICE_CONNECTION, etc.,  they appear to be set during the pipeline run—possibly for security reasons.
As for cloud, regions, and variables provided by user, we should be able to remove them in almost all scenarios.

pypeline/resource/setup.py
vittoriasalim: For validate_owner_info() function is called in the Layout during the resource validation step.

pypeline/resource/setup.py
vittoriasalim: > to record this value, which will be used in later steps. I don't think we can remove it.

Yes you are right!

pypeline/cloud/azure.py
```diff
@@ -1,99 +1,80 @@
...
     credential_type: CredentialType = CredentialType.SERVICE_CONNECTION
     azure_service_connection: str = os.getenv("AZURE_SERVICE_CONNECTION")
     azure_mi_client_id: str = os.getenv("AZURE_MI_CLIENT_ID")
```
wonderyl, line 18: Now we get the AZURE_MI_CLIENT_ID from the env we run Pypeline script. Obviously, it doesn't have it. I'm not 100% sure where is AZURE_MI_CLIENT_ID set, but I suspect it's from Azure pipeline, so we need to get the env var in the yaml file

pypeline/terraform/terraform.py
```diff
@@ -0,0 +1,187 @@
...
+            echo "Owner: $owner"
+            echo "##vso[task.setvariable variable=OWNER]$owner"
+            """
+        ).strip(),
+        condition="ne(variables['SKIP_RESOURCE_MANAGEMENT'], 'true')",
+    )
+
+
+@dataclass
+class Terraform:
```
wonderyl, line 128: Should it inherit from the Resource class?

pypeline/benchmark.py
```diff
@@ -1,67 +1,61 @@
...
     def get_jobs(self) -> list[Job]:
-   
+
         setup = Job(
             job="setup",
             display_name="Setup resources",
             steps=self.setup.setup()
             + self.cloud.login()
             + [step for r in self.resources for step in r.setup()]
+            + self.terraform.setup()
```
wonderyl, line 44: You should model Terraform as a resource and add it to self.resources.

pypeline/terraform/terraform.py
wonderyl: Actually Terraform is not a resource, it's a tool. a AKS cluster is a resource, a resource group is a resource.

pypeline/benchmark.py
```diff
@@ -1,67 +1,61 @@
...
+            + [
+                self.terraform.create_resource_group(),
+                self.terraform.run_command(command="version"),
```
wonderyl, line 47: It's better to have version, init, apply as separate functions or at least make command an enum, so that it doesn't take arbitrary string.

pypeline/resource/setup.py
wonderyl: The commend is on line 111. I was asking for `execute_tests`

vittoriasalim: Right, it should have been removed. I was testing smthg and did remove it.

PR #599, opened 2025-04-17
Adding k8s network policy benchmarking test between cilium and npm

agrawaliti:
Integrate network policy enforcement latency measurement.

Developed pipelines to compare network policy-related metrics between Azure powered by Cilium and Azure powered by CNI Overlay using Network Policy Manager.
All the configuration like nodes, pods, n
This pull request introduces significant changes to support network policy enforcement latency testing and related metrics in the cluster loader configuration. It also includes updates to enable secondary cluster usage in competitive tests. Below is a categorized breakdown of the most important changes:

### Network Policy Enforcement Latency Testing:
* Added a new deployment configuration in `deployment_template.yaml` to support network policy enforcement latency tests, including conditional labels, a new server container, and specific configurations for test pods. [[1]](diffhunk://#diff-6e902102b9ceb91638fd24417516460d679e78bec901ec817726e47a6258faa3R6-R12) [[2]](diffhunk://#diff-6e902102b9ceb91638fd24417516460d679e78bec901ec817726e47a6258faa3L21-R28) [[3]](diffhunk://#diff-6e902102b9ceb91638fd24417516460d679e78bec901ec817726e47a6258faa3L30-R73)
* Introduced new modules `net-policy-enforcement-latency.yaml` and `net-policy-metrics.yaml` to define steps for measuring network policy enforcement latency and collecting related metrics. [[1]](diffhunk://#diff-6e226ec34846ecae2bf6d6699fc018c276fe7083ba0ceba4e4b6c1ed8e6f8374R1-R55) [[2]](diffhunk://#diff-4bfff046737334c78811f229b065479bd582a79202228396537d56a3a50bdbd5R1-R122)
* Updated `load-config.yaml` to include parameters and steps for enabling and running network policy tests, such as pod creation and policy creation latency measurements. [[1]](diffhunk://#diff-3d8ff01ac3ec38f6e8991beb7c38a243602da6d6885ec24aff0c6f056feb345bL4-R5) [[2]](diffhunk://#diff-3d8ff01ac3ec38f6e8991beb7c38a243602da6d6885ec24aff0c6f056feb345bR100-R106) [[3]](diffhunk://#diff-3d8ff01ac3ec38f6e8991beb7c38a243602da6d6885ec24aff0c6f056feb345bR144-R152) [[4]](diffhunk://#diff-3d8ff01ac3ec38f6e8991beb7c38a243602da6d6885ec24aff0c6f056feb345bR176-R195) [[5]](diffhunk://#diff-3d8ff01ac3ec38f6e8991beb7c38a243602da6d6885ec24aff0c6f056feb345bR292-R304)

### Competitive Test Enhancements:
* Added `run_id_2` and `use_secondary_cluster` parameters to `competitive-test.yml` to allow tests to optionally use a secondary cluster. Updated job parameters to support these additions. [[1]](diffhunk://#diff-2f9729eea91e9c4269be4d0b0777b4a427c14c6fdf8ab18f804e1789ab6d0056R36-R38) [[2]](diffhunk://#diff-2f9729eea91e9c4269be4d0b0777b4a427c14c6fdf8ab18f804e1789ab6d0056R54-R56) [[3]](diffhunk://#diff-2f9729eea91e9c4269be4d0b0777b4a427c14c6fdf8ab18f804e1789ab6d0056R71-R76)

### Configuration Updates:
* Modified `load-config.yaml` to dynamically calculate the number of namespaces and pods per namespace based on new parameters.
* Adjusted `slo.py` to account for network tests, including a hardcoded value for nodes per namespace during network testing.

### Reconciliation Enhancements:
* Updated `reconcile-objects.yaml` to include network policy-related parameters and integration with the new deployment template. [[1]](diffhunk://#diff-3f6f706d93b5905298e0fbd3fc3c40abfcbdb78e0eecf9543a122e7c77b76b8dR17-R23) [[2]](diffhunk://#diff-3f6f706d93b5905298e0fbd3fc3c40abfcbdb78e0eecf9543a122e7c77b76b8dR68-R71)amespaces, no. of policies per namespace can be updated in pipeline.

Pipeline: https://dev.azure.com/akstelescope/telescope/_build?definitionId=41

Dashboard with new metrics: https://dataexplorer.azure.com/dashboards/e033bb3b-2cf4-4263-b41b-31597a8c4401?p-_startTime=24hours&p-_endTime=now&p-_cluster=v-cilium_network_churn_main&p-_test-type=v-default-config#5117e0aa-eb12-4f7f-b55d-6ffba1eab4ad


modules/python/clusterloader2/slo/slo.py
```diff
@@ -280,6 +314,8 @@ def main():
     parser_collect.add_argument("run_url", type=str, help="Run URL")
     parser_collect.add_argument("service_test", type=eval, choices=[True, False], default=False,
                                   help="Whether service test is running. Must be either True or False")
+    parser_collect.add_argument("network_test", type=eval, choices=[True, False], nargs='?', default=False,
```
MikeZappa87, line 317: Your branch might be old since I see type=eval present. Rebase off the latest since eval allows RCEs.

MikeZappa87 (changes requested):
RCE present

modules/python/clusterloader2/slo/slo.py
agrawaliti: done, thanks

steps/setup-tests.yml
```diff
@@ -9,17 +9,28 @@ parameters:
...
 - name: ssh_key_enabled
   type: boolean
+- name: use_secondary_cluster
+  type: boolean
+  default: false
 
 steps:
 - script: |
-    if [ -n "$RUN_ID" ]; then
+   set -eu
```
alyssa1303, line 28: This will affect all pipelines so please don't change it

steps/topology/network-churn/collect-clusterloader2.yml
```diff
@@ -0,0 +1,23 @@
...
+  default: {}
+
+steps:
+- template: /steps/engine/clusterloader2/slo/collect.yml
+  parameters:
+    cloud: ${{ parameters.cloud }}
+    engine_input: ${{ parameters.engine_input }}
+    region: ${{ parameters.regions[0] }}
+
+- script: |
```
alyssa1303, line 19: Is there a good reason for re-using cluster here? It's generally okay for 1 time feature testing or feature that requires underlay access to modify. But is that the case here? If not, then please create fresh cluster with terraform for every run as re-use clusters can lead to metric issues if the cluster goes into bad state

jobs/competitive-test.yml
```diff
@@ -33,6 +33,9 @@ parameters:
 - name: run_id
   type: string
   default: ''
+- name: run_id_2 # This cluster will be used if 'use_secondary_cluster' is true
```
alyssa1303, line 36: Please don't change this as it affects all pipelines

steps/execute-tests.yml
```diff
@@ -13,6 +13,13 @@ parameters:
   default: {}
 
 steps:
+- script: |
```
alyssa1303, line 16: Same here. Revert as this will affect all pipelines

modules/python/clusterloader2/slo/config/load-config.yaml
```diff
@@ -14,11 +15,11 @@ name: load-config
 {{$groupName := DefaultParam .CL2_GROUP_NAME "service-discovery"}}
 
 # TODO(jshr-w): This should eventually use >1 namespace.
-{{$namespaces := 1}}
+{{$namespaces := DefaultParam .CL2_NO_OF_NAMESPACES 1}}
 {{$nodes := DefaultParam .CL2_NODES 1000}}
 
 {{$operationTimeout := DefaultParam .CL2_OPERATION_TIMEOUT "15m"}}
-{{$totalPods := MultiplyInt $namespaces $nodes $podsPerNode}}
+{{$totalPods := MultiplyInt $namespaces $nodesPerNamespace $podsPerNode}}
```
jshr-w, line 22: This is going to break the existing pipelines where nodes_per_namespace is set to 100 but only 1 ns is being used. Should do this in a way that is compatible with both.

jshr-w (changes requested):
Please share the runs on our default Cluster Churn and Service Churn pipelines off this branch first so we can make sure they aren't broken.

jobs/competitive-test.yml
agrawaliti: I am working on removing this logic, as Ill create clusters using terraform, but just to understand as its an optional parameter why do you think it will affect other pipelines.

steps/execute-tests.yml
agrawaliti: is that a problem? its just collecting start time, and show that.

steps/setup-tests.yml
agrawaliti: ack

modules/python/clusterloader2/slo/config/load-config.yaml
jshr-w: Yes, that's what I said -- I didn't say work around the incorrect logic, but it's not really a "fix" if it breaks the existing pipelines. You should fix it entirely so everything works correctly.

agrawaliti: Thanks for pointing out.

steps/execute-tests.yml
MikeZappa87: As a best practice, minimize the code changes to unrelated areas. While this most likely has 0 impact, we should scope changes to impact just our code.

PR #614, opened 2025-05-01
Update CAS Profile and default cooldown period to 120 seconds

sumanthreddy29:
1. Update CAS profile
2. Add 120 seconds wait time before we scale down and before the next iteration to run.


modules/python/clusterloader2/autoscale/config/config.yaml
```diff
@@ -11,7 +11,7 @@
 {{$scaleDownTimeout := DefaultParam .CL2_SCALE_DOWN_TIMEOUT "10m"}}
 {{$refreshInterval := DefaultParam .CL2_REFRESH_INTERVAL "5s"}}
 {{$loopCount := DefaultParam .CL2_LOOP_COUNT 1}}
-{{$coolDownTime := DefaultParam .CL2_COOLDOWN_TIME "30s"}}
+{{$coolDownTime := DefaultParam .CL2_COOLDOWN_TIME "120s"}}
```
alyssa1303, line 14: Why don't make it a parameter from input instead of update default?

alyssa1303: This is actually used for both create and delete. So if you just want to update for delete, you need to split this into 2 parameters

sumanthreddy29: We plan on using same `coolDownTime` for both CAS and NAP for all nodes.  The reason I didn't use parameters is we are not planning to customize the wait time for different nodes, so I updated the default value.

sumanthreddy29: We want to use 2mins for both create and Delete.

sumanthreddy29: Scale up > Wait 2 mins -> Scale down -> Wait 2 mins -> 2nd Iteration - Scaleup

alyssa1303 (approved):

PR #637, opened 2025-05-08
Skip Scheduled runs from Non-main branch

sumanthreddy29:
1. Update Competitive job template to check if scheduled runs are from main branch or not.

### DIFF
--- a/jobs/competitive-test.yml
+++ b/jobs/competitive-test.yml
@@ -50,57 +50,64 @@ parameters:
   default: true
 
 jobs:
-- job: ${{ parameters.cloud }}
-  strategy:
-    maxParallel: ${{ parameters.max_parallel }}
-    matrix:
-      ${{ parameters.matrix }}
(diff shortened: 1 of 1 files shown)


jobs/competitive-test.yml
```diff
@@ -56,6 +56,7 @@ jobs:
     matrix:
       ${{ parameters.matrix }}
   timeoutInMinutes: ${{ parameters.timeout_in_minutes }}
+  condition: or(eq(variables['Build.Reason'], 'Manual'), and(eq(variables['Build.Reason'], 'Schedule'), eq(variables['Build.SourceBranchName'], 'main')))
```
anson627, line 59: we still want to allow Manual, right?

anson627 (approved):

jobs/competitive-test.yml
sumanthreddy29: Yes, We allow manual runs here.

alyssa1303 (approved):

PR #639, opened 2025-05-09
Add yaml files for pod-to-pod test

alyssa1303:


modules/kustomize/iperf3/base/iperf3.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: Deployment
+metadata:
+  name: iperf3-deployment
+  namespace: default
+spec:
+  replicas: 1
+  template:
+    spec:
+      containers:
+        - image: telescope.azurecr.io/perf-eval/iperf3.18:v20250415
```
sumanthreddy29, line 11: Can we add an option to take this param from pipeline instead of hardcoding it in this yaml file?

modules/kustomize/iperf3/base/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - iperf3.yaml
```
sumanthreddy29, line 5: We can add images here which get replaced in the yaml file.

``` images:
 - name: image
   newName: my.docker.registry.com/my-project/my-app
   newTag: test

We can export our image value in the validate may be and which will replace the image in iperf3.yaml

File: modules/kustomize/iperf3/base/iperf3.yaml (hunk 1 above)
Comment (sumanthreddy29) [reply modules/kustomize/iperf3/base/iperf3.yaml:11]: In future if we need to change image tag, we have to update all base yaml files with new version.

Comment (alyssa1303) [reply modules/kustomize/iperf3/base/iperf3.yaml:11]: There's only 1 base which is this file, so update image in the base will update the image for all variants.

File: modules/kustomize/iperf3/base/kustomization.yaml (hunk 2 above)
Comment (alyssa1303) [reply modules/kustomize/iperf3/base/kustomization.yaml:5]: Yes we can but we should use same image for all iperf3 tests, so ideally update 1 file at the base should work instead of customizing here.

**Review by sumanthreddy29 — APPROVED** [review]

File: steps/engine/iperf3/validate.yml [hunk 3]
```diff
@@ -0,0 +1,18 @@
...
+  cluster_cli_context:
+    type: string
+  cluster_srv_context:
+    type: string
+
+steps:
+  - script: |
+      set -eo pipefail
+      PYTHONPATH=$PYTHONPATH:$(pwd) python3 $PYTHON_SCRIPT_FILE validate \
+      --cluster_cli_context "$CLUSTER_CLI_CONTEXT" \
```
anson627, line 11: client_context and server_context to be clear, cli can also mean command line interface

modules/kustomize/iperf3/overlays/client-hostnetwork-true/kustomization.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+patches:
+  - patch: |-
+      - op: add
+        path: /spec/template/spec/hostNetwork
+        value: true
+      - op: add
+        path: /spec/template/spec/nodeSelector
+        value:
+          app: client
```
anson627, line 16: can we select node by zone?

modules/kustomize/iperf3/overlays/client-hostnetwork-true/kustomization.yaml
anson627, line 1: this doesn't look like scalable: client or server, hostnetwork: true, false, what if we add another dimension?

can we rename overlays? these are deployments

pipelines/perf-eval/Network Benchmark/pod-to-pod-diff-node-same-cluster-vnet.yml
```diff
@@ -0,0 +1,100 @@
...
+  - cron: "0 23 * * 2"
+    displayName: "11:00 PM Every Tuesday"
+    branches:
+      include:
+        - main
+    always: true
+
+variables:
+  SCENARIO_TYPE: perf-eval
+  SCENARIO_NAME: pod-diff-node-vnet
```
anson627, line 19: different node same zone, right?

steps/engine/iperf3/configure.yml
anson627, line 1: rename to configue-pod?

modules/kustomize/iperf3/overlays/client-hostnetwork-true/kustomization.yaml
alyssa1303: I've update hostNetwork to be a component and only have 2 overlays, 1 for server and 1 for client. If we have a new dimension, we can make it a component and build it on the fly so we don't have to create all the possible combinations of final setup

modules/kustomize/iperf3/overlays/client-hostnetwork-true/kustomization.yaml
alyssa1303: Zone is decided at cluster setup level so use label to distinguish between client and server since cluster setup can have nodes in same zone. Selecting by zone only can lead to both pods landing on same node

steps/engine/iperf3/configure.yml
alyssa1303: done

steps/engine/iperf3/validate.yml
alyssa1303: updated

PR #642, opened 2025-05-11
Decouple run command in terraform by implementing cloud specific logic in Cloud class

vittoriasalim:
- Move cloud specific logic to cloud class

### DIFF
--- a/pypeline/cloud/azure.py
+++ b/pypeline/cloud/azure.py
@@ -131,14 +131,12 @@ class Azure(Cloud):
             ),
         }
 
-    def generate_tf_error_handler(self, command: str) -> str:
-        if command == "apply":
-            return dedent(
-                """echo "Deleting resources and removing state file before retrying"
-                    ids=$(az resource list --location $region --resource-group $RUN_ID --query [*].id -o tsv)
-                    az resource delete --ids $ids --verbose
-                    rm -r terraform.tfstate.d/$region
-                """
-            ).strip()
-        else:
-            return ""
+    def delete_resource_group(self) -> str:
+        return dedent(
+            """
+            echo "Deleting resources and removing state file before retrying"
+            ids=$(az resource list --location $region --resource-group $RUN_ID --query [*].id -o tsv)
+            az resource delete --ids $ids --verbose
+            rm -r terraform.tfstate.d/$region
+            """
+        ).strip("")
--- a/pypeline/cloud/azure.py
+++ b/pypeline/cloud/azure.py
@@ -130,3 +130,15 @@ class Azure(Cloud):
                 "user_node_pool", "$USER_NODE_POOL"
             ),
         }
+
+    def generate_tf_error_handler(self, command: str) -> str:
+        if command == "apply":
+            return dedent(
+                """echo "Deleting resources and removing state file before retrying"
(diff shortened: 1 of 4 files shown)


pypeline/terraform/terraform.py
```diff
@@ -192,6 +290,30 @@ def setup(self) -> list[Step]:
             set_input_variables(self.cloud, self.regions, self.input_variables),
             get_deletion_info(self.regions[0]),
             create_resource_group(self.regions[0], self.cloud.provider.value),
+            run_command(
+                "version",
```
wonderyl, line 294: Could you make the commands enum to avoid typo?

pypeline/cloud/azure.py
```diff
@@ -130,3 +130,15 @@ def generate_input_variables(self, region: str, input_variables: dict) -> dict:
                 "user_node_pool", "$USER_NODE_POOL"
             ),
         }
+
+    def generate_tf_error_handler(self, command: str) -> str:
+        if command == "apply":
```
wonderyl, line 135: I wonder if it's possible to move the `command == "apply"` condition into the terraform class. When this is in the Cloud class, the Cloud class is more coupled with the Terraform class and its implementation.

pypeline/cloud/azure.py
```diff
@@ -130,3 +130,15 @@ def generate_input_variables(self, region: str, input_variables: dict) -> dict:
...
+            return dedent(
+                """echo "Deleting resources and removing state file before retrying"
+                    ids=$(az resource list --location $region --resource-group $RUN_ID --query [*].id -o tsv)
```
wonderyl, line 138: Why there's an extra white space before line 138-140?

pypeline/terraform/terraform.py
```diff
@@ -168,8 +168,106 @@ def set_input_variables(
     )
 
 
+def generate_workspace_script() -> str:
+    return dedent(
+        """if terraform workspace list | grep -q "$region"; then
+              terraform workspace select $region
```
wonderyl, line 174: Ditto, lines are not aligned

pypeline/terraform/terraform.py
```diff
@@ -192,6 +290,30 @@ def setup(self) -> list[Step]:
...
             create_resource_group(self.regions[0], self.cloud.provider.value),
+            run_command(
+                "version",
+                self.arguments,
+                self.regions,
+                self.cloud,
+                self.retry_attempt_count,
+                self.credential_type,
+            ),
+            run_command(
```
wonderyl, line 301: Looks like you pass similar arguments to run_command, would it be easier if you make run_command a member funtion instead, then you don't need to pass any member of self as arguments.

pypeline/terraform/terraform.py
```diff
@@ -290,34 +270,43 @@ def setup(self) -> list[Step]:
...
-            ),
-            run_command(
-                "apply",
-                self.arguments,
-                self.regions,
-                self.cloud,
-                self.retry_attempt_count,
-                self.credential_type,
-            ),
+            self.run_command(TerraformCommand.INIT),
```
wonderyl, line 273: Is the order right? it used to call version first.

vittoriasalim: My bad, updated

wonderyl (approved):

PR #643, opened 2025-05-12
Move resource group from terraform to cloud class 

vittoriasalim:
- Move creating resource group to cloud class for decoupling 
- To not expose yaml condition in checking if cloud is azure, aws or gcp

Previously condition was checking if azure == azure 
with **condition=and(eq('{cloud}', 'azure')**
By moving to cloud class this can be removed

### DIFF
--- a/pypeline/cloud/azure.py
+++ b/pypeline/cloud/azure.py
@@ -131,7 +131,7 @@ class Azure(Cloud):
             ),
         }
 
-    def create_resource_group(self, region: str) -> Script:
+    def create_resource_group(self, region: str) -> str:
         return dedent(
             f"""
             set -eu
--- a/pypeline/cloud/azure.py
+++ b/pypeline/cloud/azure.py
@@ -132,18 +132,14 @@ class Azure(Cloud):
         }
 
     def create_resource_group(self, region: str) -> Script:
-        return Script(
-            display_name="Create Resource Group",
-            script=dedent(
-                f"""
-                set -eu
-                echo "Create resource group $RUN_ID in region {region}"
-                az group create --name $RUN_ID --location {region} \\
-                --tags "run_id=$RUN_ID" "scenario=${{SCENARIO_TYPE}}-${{SCENARIO_NAME}}" "owner=${{OWNER}}" "creation_date=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" "deletion_due_time=${{DELETION_DUE_TIME}}" "SkipAKSCluster=1"
-                """
-            ).strip(),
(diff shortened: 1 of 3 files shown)


pypeline/cloud/azure.py
```diff
@@ -131,6 +131,20 @@ def generate_input_variables(self, region: str, input_variables: dict) -> dict:
...
+            display_name="Create Resource Group",
+            script=dedent(
+                f"""
+                set -eu
+                echo "Create resource group $RUN_ID in region {region}"
+                az group create --name $RUN_ID --location {region} \\
+                --tags "run_id=$RUN_ID" "scenario=${{SCENARIO_TYPE}}-${{SCENARIO_NAME}}" "owner=${{OWNER}}" "creation_date=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" "deletion_due_time=${{DELETION_DUE_TIME}}" "SkipAKSCluster=1"
+                """
+            ).strip(),
+            condition="ne(variables['SKIP_RESOURCE_MANAGEMENT'], 'true')",
```
wonderyl, line 145: This requires the cloud provider to set condition, which is not really in the responsibility of a cloud provider. 
You can follow the pattern of `delete_resource_group` where only the command is returned.

pypeline/terraform/terraform.py
```diff
@@ -129,22 +129,6 @@ def get_deletion_info(region: str) -> Script:
...
-        display_name="Create Resource Group",
-        script=dedent(
-            f"""
-            set -eu
-            echo "Create resource group $RUN_ID in region {region}"
-            az group create --name $RUN_ID --location {region} \
-            --tags "run_id=$RUN_ID" "scenario=${{SCENARIO_TYPE}}-${{SCENARIO_NAME}}" "owner=${{OWNER}}" "creation_date=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" "deletion_due_time=${{DELETION_DUE_TIME}}" "SkipAKSCluster=1"
-            """
-        ).strip(),
-        condition=f"and(eq('{cloud}', 'azure'), ne(variables['SKIP_RESOURCE_MANAGEMENT'], 'true'))",
```
wonderyl, line 144: How is `eq('{cloud}', 'azure')` implemented?

pypeline/cloud/azure.py
```diff
@@ -131,6 +131,16 @@ def generate_input_variables(self, region: str, input_variables: dict) -> dict:
             ),
         }
 
+    def create_resource_group(self, region: str) -> Script:
```
wonderyl, line 134: return type should be str

wonderyl (approved):

PR #646, opened 2025-05-12
Create pipeline to run ACNS scale test

alexcastilio:
Create pipeline to run ACNS with all its corresponding topology and scenarios


modules/python/clusterloader2/utils.py
```diff
@@ -19,17 +19,20 @@
...
 def run_cl2_command(kubeconfig, cl2_image, cl2_config_dir, cl2_report_dir, provider, cl2_config_file="config.yaml", overrides=False, enable_prometheus=False, tear_down_prometheus=True,
-                    enable_exec_service=False, scrape_kubelets=False, scrape_containerd=False):
+                    enable_exec_service=False, scrape_kubelets=False, scrape_containerd=False, scrape_ksm=False, scrape_metrics_server=False):
     docker_client = DockerClient()
 
     command=f"""--provider={provider} --v=2
 --enable-exec-service={enable_exec_service}
 --enable-prometheus-server={enable_prometheus}
 --prometheus-scrape-kubelets={scrape_kubelets}
+--prometheus-scrape-master-kubelets={scrape_kubelets}
```
alyssa1303, line 29: Do we have master kubelet to scrape? Might be better to have a separate parameter for this and set it to False to avoid breaking existing tests.

modules/python/clusterloader2/scale/config/modules/kube-state-metrics/clusterrole.yaml
alyssa1303, line 1: I'm a bit confused why this folder is needed here when it's already in [perf-tests](https://github.com/Azure/perf-tests/tree/main/clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics)?

modules/python/clusterloader2/scale/config/modules/node-exporter/clusterrole.yaml
alyssa1303, line 1: Same question about this `node-exporter` folder? I believe they should come with the [cl2 image](https://github.com/Azure/perf-tests/blob/main/clusterloader2/pkg/prometheus/manifests/exporters/node_exporter/node-exporter.yaml). If somethins is missing, the better place to put them is in perf-tests instead of here so other tests can re-use the modules. Ideally, once we enable the flag, these should be applied automatically from CL2

modules/python/clusterloader2/utils.py
alexcastilio: When --prometheus-scrape-master-kubelets is set to true, 2 endpoints for kubelet are added to master service monitor deployed by CL2. I'm collecting memory, cpu for several pods and api server. I'm adding a separate parameter for it

modules/python/clusterloader2/scale/config/modules/kube-state-metrics/clusterrole.yaml
alexcastilio: I have done it this way, to move quicker during development as it was not working out of the box initially. PR raised for it: https://github.com/Azure/perf-tests/pull/37

modules/python/clusterloader2/scale/config/modules/node-exporter/clusterrole.yaml
alexcastilio: Node Exporter is a bit trickier. The way node exporter is deployed currently with CL2 is to SSH access to nodes and deploying a pod manifest in each node: https://github.com/Azure/perf-tests/blob/bb3ffbafcfe43676bffb0124a989389f8e427cbd/clusterloader2/pkg/prometheus/prometheus.go#L528C1-L564C1

To change to a daemonset-based approach would require a bigger piece of change in perf-tests repo with potentially some discussions about implications of changing this approach.

I know this is not ideal, but this was a way to make it work quickly to complete development of test pipeline. If you agree, I would prefer to keep this part as it is now, and once this is changed in perf-tests, we delete this part accordingly. Is that ok?

modules/python/clusterloader2/utils.py
alyssa1303, line 1: Let do this in a separate PR with some testing

modules/python/clusterloader2/scale/config/modules/measurements.yaml
alyssa1303, line 1: Rename this file to clearly state what metrics are being measured here like other files, then move to `measurements` folder

modules/python/clusterloader2/scale/config/modules/node-exporter.yaml
alyssa1303, line 1: This looks more like setup file then measurement file. Let's rename the file to correctly reflect what it does. Also, what is the exact metrics being looked at here?

modules/python/clusterloader2/scale/config/modules/cilium-measurements.yaml
alyssa1303, line 1: Move to folder called `measurements`

steps/topology/observability/validate-resources.yml
```diff
@@ -0,0 +1,16 @@
...
+    type: object
+
+steps:
+  - template: /steps/cloud/${{ parameters.cloud }}/update-kubeconfig.yml
+    parameters:
+      role: slo
+      region: ${{ parameters.regions[0] }}
+# - template: /steps/engine/clusterloader2/slo/validate.yml
+#   parameters:
+#     desired_nodes: $(node_count)
```
alyssa1303, line 16: If not used, please remove

steps/topology/observability/collect-clusterloader2.yml
```diff
@@ -0,0 +1,22 @@
...
+- template: /steps/engine/clusterloader2/scale/collect.yml
+  parameters:
+    cloud: ${{ parameters.cloud }}
+    engine_input: ${{ parameters.engine_input }}
+    region: ${{ parameters.regions[0] }}
+- script: |
+    run_id=$(Build.BuildId)-$(System.JobId)
+    echo "Run ID: $run_id"
+    echo "##vso[task.setvariable variable=RUN_ID]$run_id"
+  displayName: "Set unique Run ID before publish"
```
alyssa1303, line 22: Why is this step needed?

alyssa1303 (commented):
Can we rename the folder from `scale` to `acns` or something to reflect what it's trying to test here?

scenarios/perf-eval/acns-azurecni-overlay-cilium/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,79 @@
...
+deletion_delay = "20h"
+owner          = "aks"
+
+aks_cli_config_list = [
+  {
+    role               = "slo"
+    aks_name           = "telescope-acns-scale-test"
+    sku_tier           = "standard"
+    kubernetes_version = "1.32"
+    sku_tier           = "Standard"
```
alyssa1303, line 12: Duplicate? Remove

scenarios/perf-eval/acns-azurecni-overlay/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,75 @@
...
+deletion_delay = "20h"
+owner          = "aks"
+
+aks_cli_config_list = [
+  {
+    role               = "slo"
+    aks_name           = "telescope-acns-scale-test"
+    sku_tier           = "standard"
+    kubernetes_version = "1.32"
+    sku_tier           = "Standard"
```
alyssa1303, line 12: Remove

scenarios/perf-eval/non-acns-azurecni-overlay/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,71 @@
...
+deletion_delay = "20h"
+owner          = "aks"
+
+aks_cli_config_list = [
+  {
+    role               = "slo"
+    aks_name           = "telescope-acns-scale-test"
+    sku_tier           = "standard"
+    kubernetes_version = "1.32"
+    sku_tier           = "Standard"
```
alyssa1303, line 12: Remove

scenarios/perf-eval/acns-azurecni-overlay-cilium/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,79 @@
+scenario_type  = "perf-eval"
+scenario_name  = "retina-oss-standard-basic"
```
alyssa1303, line 2: scenario_name should match folder name. This should be `acns-azurecni-overlay-cilium`

scenarios/perf-eval/acns-azurecni-overlay/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,75 @@
+scenario_type  = "perf-eval"
+scenario_name  = "acns-azurecni-overlay"
```
alyssa1303, line 2: scenario_name should match folder name.

scenarios/perf-eval/non-acns-azurecni-overlay/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,71 @@
+scenario_type  = "perf-eval"
+scenario_name  = "acns-azurecni-overlay"
```
alyssa1303, line 2: scenario_name should match folder name.

modules/python/clusterloader2/scale/config/modules/kube-state-metrics.yaml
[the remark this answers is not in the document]
alexcastilio: This is only deploying ksm. Metrics are collected under `measurements/` folder. Regarding ksm I'm getting kube_pod_container_status_restarts_total for AMA Logs, retina and cilium

steps/topology/observability/collect-clusterloader2.yml
alexcastilio: I saw this in some other tests (topologies cilium-usercluster-autoscale, cilium-usercluster, network-load, network-policy-scale) to set a unique Run ID so I can correlated data published to corresponding execution when it's published

modules/python/clusterloader2/utils.py
alexcastilio: PR for this: https://github.com/Azure/telescope/pull/649
Please let me know if you need something else from my side on this, as I will need this PR merged to be able to continue running the pipeline

modules/python/clusterloader2/scale/config/modules/node-exporter.yaml
alexcastilio: This is to deploy node exporter components. What name you suggest? Metrics collected are stored in files under `measurements` folder

modules/python/clusterloader2/scale/config/config.yaml
```diff
@@ -0,0 +1,95 @@
...
+          duration: 5m
+
+  - module:
+      path: /modules/scale-test.yaml
+      params:
+        action: start
+        group: {{$groupName}}
+        createRetinaNetworkFlowLogs: true
+        labelTrafficPods: {{$labelTrafficPods}}
+    {{end}}
```
alyssa1303, line 88: This is redundant. Why don't you use `generateRetinaNetworkFlowLogs` in `createRetinaNetworkFlowLogs` and let scale-test handle the rest? This should suffice:
```
  - module:
      path: /modules/scale-test.yaml
      params:
        action: start
        group: {{$groupName}}
        createRetinaNetworkFlowLogs: {{$generateRetinaNetworkFlowLogs}}
        labelTrafficPods: {{$labelTrafficPods}}
```
Even better, just use the same name for parameters instead of 2 different names to do the same thing as they're both boolean value.
```
  - module:
      path: /modules/scale-test.yaml
      params:
        action: start
        group: {{$groupName}}
        createRetinaNetworkFlowLogs: {{$createRetinaNetworkFlowLogs}}
        labelTrafficPods: {{$labelTrafficPods}}
```

modules/python/clusterloader2/scale/config/modules/reconcile-objects.yaml
```diff
@@ -0,0 +1,362 @@
...
+    - ""
+    replicasPerNamespace: 0
+    tuningSet: {{$tuningSet}}
+    objectBundle:
+    - basename: retina-network-flow-log
+      objectTemplatePath: modules/pfl/retinanetworkflowlog.yaml
+      templateFillMap:
+        Group: {{.Group}}
+        namespaces: {{$namespaces}}
+{{end}}
```
alyssa1303, line 362: If you want to make reconcile-objects into a template, then you should allow customization of `replicasPerNamespace` so for create action, you specific the desired replica, and for delete action, you specific 0 like this [file](https://github.com/Azure/telescope/blob/main/modules/python/clusterloader2/slo/config/modules/reconcile-objects.yaml). This part can then be removed and in the scale-test. it should call reconcile-object with delete action.

steps/topology/observability/collect-clusterloader2.yml
alexcastilio: The difference is that when I'm manually running test on the same cluster that I have created by skipping infra/resource management the RUN_ID does not change, because of the cluster tag for run_id. So this will override the previous one, otherwise all runs would have the same run_id

modules/python/clusterloader2/scale/config/config.yaml
alexcastilio: Because they are 2 separate executions of actions in the cluster with 2 separate sets of measurements. The first always happens. The second one happens based on the flag set to true. Metrics of the second execution have a suffix appended.

modules/python/clusterloader2/scale/config/modules/reconcile-objects.yaml
alexcastilio: You are right. It started as a template following other pipelines, but it turned to be the sequence of steps of a test round. I have renamed it to `scale-test.yaml`. I hope it's clearer now

modules/python/clusterloader2/scale/config/modules/scale-test.yaml
```diff
@@ -0,0 +1,120 @@
...
+      params:
+        action: start
+        group: {{$groupName}}
+        suffix: "{{$suffix}}"
+
+  - module:
+      path: /modules/reconcile-objects.yaml
+      params:
+        actionName: "create"
+        tuningSet: DeploymentCreateQps
```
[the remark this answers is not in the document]
alexcastilio: I have removed actionName

PR #649, opened 2025-05-13
Add metric server and kube state metric flag

alexcastilio:


modules/python/clusterloader2/utils.py
```diff
@@ -16,17 +16,21 @@
...
-                    enable_exec_service=False, scrape_kubelets=False, scrape_containerd=False):
+                    enable_exec_service=False, scrape_kubelets=False, scrape_master_kubelets=False,
+                    scrape_containerd=False, scrape_ksm=False, scrape_metrics_server=False):
     docker_client = DockerClient()
 
     command=f"""--provider={provider} --v=2
 --enable-exec-service={enable_exec_service}
 --enable-prometheus-server={enable_prometheus}
 --prometheus-scrape-kubelets={scrape_kubelets}
+--prometheus-scrape-master-kubelets={scrape_master_kubelets}
```
alyssa1303, line 27: Why is this one needed when we don't have mster nodes on overlay?

alexcastilio: I need that flag to scrape cadvisor: https://github.com/Azure/perf-tests/blob/f1933d032aef81afd06153fbcb778509d9e54874/clusterloader2/pkg/prometheus/manifests/master-ip/master-serviceMonitor.yaml#L42C3-L60C10

alyssa1303: Scrape kubelet already does that https://github.com/Azure/perf-tests/blob/f1933d032aef81afd06153fbcb778509d9e54874/clusterloader2/pkg/prometheus/manifests/default/prometheus-serviceMonitorKubelet.yaml#L24. I don't think it's needed here since we don't have master kubelet in overlay

alexcastilio: Ok. maybe I did something wrong when I started looking at this part. Let's try without that. And if needed I'll raise another PR to add it.

alyssa1303 (approved):

PR #652, opened 2025-05-15
Add new stage and update scenario names and labels for pod to pod test

alyssa1303:
Summary:
- Update scenario name to match setup more clearly
- Include new stage for v3 and v5 from internal
- Update node labels so it can be used for case of deploying pods on same node


pipelines/perf-eval/Network Benchmark/pod-to-pod-diff-node-same-zone-same-cluster.yml
```diff
@@ -30,9 +36,43 @@ parameters:
...
+                    k8s_machine_type: ${{ machine_type }}
+                    network_dataplane: ${{ dataplane }}
+                    host_network: ${{ network }}
+                    configure_mtu: "false"
+          max_parallel: 2
+          timeout_in_minutes: 120
+          credential_type: service_connection
+          ssh_key_enabled: false
+
   - stage: azure_eastus2_new_sku
```
sumanthreddy29, line 76: Can we use sku name instead of new sku here and for other stage names, so that It's easy for us to identify which one is v3, v5, v6?

alyssa1303: Updated

sumanthreddy29 (approved):

PR #661, opened 2025-05-20
Implement CLI Argument Parser for KWOK Topology Invocation

vittoriasalim:
Implement an argument parser so that KWOK can be invoked in the topology.

**Create 2 virtual nodes using the latest KWOK release:**

> python -m kwok.kwok --mode node --action create --node-count 2

**Validate that 2 KWOK nodes are present**

> python -m kwok.kwok --mode node --action validate --node-count 2

**Tear down (delete) 2 KWOK nodes**

> python -m kwok.kwok --mode node --action tear_down --node-count 2

**Create nodes with a specific KWOK release and enable metrics:**

> python -m kwok.kwok --mode node --action create --node-count 2 --kwok-release v0.7.0 --enable-metrics

**Use a custom node manifest:**

> python -m kwok.kwok --mode node --action create --node-manifest-path path/to/custom-node.yaml

### DIFF
--- a/modules/python/kwok/kwok.py
+++ b/modules/python/kwok/kwok.py
@@ -1,6 +1,8 @@
 import subprocess
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+import argparse
+import sys
 
 import requests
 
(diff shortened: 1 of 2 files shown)


modules/python/kwok/kwok.py
```diff
@@ -143,4 +144,68 @@ def tear_down(self):
...
+    parser.add_argument(
+        "--node-manifest-path",
+        type=str,
+        default="kwok/config/kwok-node.yaml",
+        help="Path to the node manifest YAML template.",
+    )
+    parser.add_argument(
+        "--kwok-release",
+        type=str,
+        default=None,
```
alyssa1303, line 172: Should this be "latest"?

vittoriasalim: I keep it as None as the logic work as following:
If self.kwok_release already has a truthy value (i.e., it's not None, False, 0, '', etc.), keep it. Otherwise, assign it the result of self.fetch_latest_release().
https://github.com/Azure/telescope/blob/6da41447a8b081d0da656924f4dca207ecc090cf/modules/python/kwok/kwok.py#L58C11-L58C81

alyssa1303: Can we use "" instead of None to align with the type of str?

alyssa1303 (approved):

PR #666, opened 2025-05-21
increase csi attach test frequency

umagnus:
increase csi attach test frequency, start with running the tests once everyday instead of every other day


feiskyer (approved):
looks good to me

sumanthreddy29 (approved):

PR #670, opened 2025-05-27
Remove create resource group from cloud module since it is only relevant to Azure

vittoriasalim:
Asked Lei about this, there is no resource group at AWS and GCP, since create_resource_group is only relevant to azure. This PR is **to remove create_resource_group function from Cloud class**

### DIFF
--- a/pypeline/benchmark.py
+++ b/pypeline/benchmark.py
@@ -52,10 +52,6 @@ class Cloud(ABC):
     def delete_resource_group(self) -> str:
         pass
 
-    @abstractmethod
-    def create_resource_group(self) -> str:
-        pass
-
 
 @dataclass
 class Layout:
--- a/pypeline/terraform/terraform.py
+++ b/pypeline/terraform/terraform.py
@@ -233,9 +233,14 @@ def generate_apply_or_destroy_script(
 
 
 def create_resource_group(cloud: Cloud, region: str) -> Script:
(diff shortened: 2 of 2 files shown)


pypeline/benchmark.py
```diff
@@ -52,10 +52,6 @@ def generate_input_variables(self, region: str, input_variables: dict) -> dict:
     def delete_resource_group(self) -> str:
```
wonderyl, line 52: What about delete?

vittoriasalim: it is in the next PR hahaha

pypeline/terraform/terraform.py
```diff
@@ -233,9 +233,14 @@ def generate_apply_or_destroy_script(
 
 
 def create_resource_group(cloud: Cloud, region: str) -> Script:
+    if cloud.provider.value != CloudProvider.AZURE:
```
wonderyl, line 236: I think you can use `if cloud.provider != CloudProvider.AZURE:`

pypeline/terraform/terraform.py
```diff
@@ -233,9 +233,14 @@ def generate_apply_or_destroy_script(
...
+        script_str = cloud.create_resource_group(region)
```
wonderyl, line 237: Since create_resource_group is removed from the Cloud class, you can't call it anymore, right?

PR #671, opened 2025-05-27
Rename delete_resource_group to delete_resources, since resource group is not relevant to cloud module (aws and gcp do not have resource group)

vittoriasalim:
- Rename _**delete_resource_group**_ to _**delete_resources**_ since **_delete_resource_group_** is not relevant to the cloud module, as GCP and AWS do not have the concept of a resource group.

- TODO: The delete_resource_group function will be added to the Azure class only.

### DIFF
--- a/pypeline/benchmark.py
+++ b/pypeline/benchmark.py
@@ -49,7 +49,7 @@ class Cloud(ABC):
         pass
 
     @abstractmethod
-    def delete_resource_group(self) -> str:
+    def delete_resources(self) -> str:
         pass
 
     @abstractmethod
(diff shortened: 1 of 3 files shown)


pypeline/benchmark.py
```diff
@@ -49,7 +49,7 @@ def generate_input_variables(self, region: str, input_variables: dict) -> dict:
         pass
 
     @abstractmethod
-    def delete_resource_group(self) -> str:
+    def delete_resources(self) -> str:
```
wonderyl, line 52: It'll be cleaner if each resource can delete themselves, rather than let the Cloud to delete them.

PR #668, opened 2025-05-22
Skip deleting resource for debugging

wonderyl:
When we debug a new pipeline and it fails, the cluster is deleted before we have a chance to investigate.
This PR adds a variable "SKIP_RESOURCE_DELETION". When set, it the pipeline to skips the deletion of resources in manual runs.
The owner of a pipeline run is responsible to delete the resource manually when it's skipped.

To use it, one can set the variable in the pipeline yaml.

```
variables:
    SKIP_RESOURCE_DELETION: true
```

This is tested in https://dev.azure.com/akstelescope/telescope/_build/results?buildId=25279&view=logs&j=de223fca-0b0d-5d04-db32-a8c4d0bcfd88&t=c58ccdb9-3032-53d1-f00a-a15c1c92ae2e

PR #690, opened 2025-06-03
Model resource group class

vittoriasalim:
- This PR is to create resource group class in handling resource group creation and deletion
- Model resource group class to decouple cloud module

Edit #2 
- Make owner and deletion delay as field property 
- Remove get deletion delay and owner step from terraform
- Move ResourceGroup to resources list, make it as a resource of its own
- Remove condition in checking cloud provider, it is safe to assume resource group only belong to azure
- Remove parameters from setup() in resource group and pass from dataclass property
- TODO: Make PR for skip delete resource management , will be done in another PR

Edit #3 
- remove deletion_delay and owner from terraform.py

Edit #4 
- replace region with azure cloud regions
- replace cloud string with cloud provider for ssh

Pipeline: [Pipelines - Run 20250604.6 logs](https://dev.azure.com/akstelescope/telescope/_build/results?buildId=25475&view=logs&j=b494c9ec-15f6-565b-9772-130ae1974e1c&t=b494c9ec-15f6-565b-9772-130ae1974e1c)

### DIFF
--- a/pypeline/pipelines/controller/job_scheduling.py
+++ b/pypeline/pipelines/controller/job_scheduling.py
@@ -1,5 +1,4 @@
 import os
-from terraform.resource_group import ResourceGroup
 from resource.python3 import Python3
 from resource.setup import Setup
 from resource.ssh import SSH
@@ -7,14 +6,13 @@ from resource.ssh import SSH
 from benchmark import Benchmark, Layout
 from cloud.azure import Azure
 from engine.clusterloader2 import ClusterLoader2
+from terraform.resource_group import ResourceGroup
 from terraform.terraform import Terraform
 
 
 def main():
     # TODO : Refactor, make function to generate layout
-    azure_eastus2 = Azure(
-        region="eastus2"
-    )
+    azure_eastus2 = Azure(region="eastus2")
     job_scheduling = Benchmark(
         name="job_scheduling",
         layouts=[
@@ -28,9 +26,10 @@ def main():
                         cloud=azure_eastus2,
                         regions=["eastus2"],
                         scenario_name="job-scheduling",
+                        deletion_delay=2,
                     ),
                     Python3(),
-                    SSH(cloud="azure")
+                    SSH(cloud="azure"),
                 ],
                 engine=ClusterLoader2(),
             )
--- a/pypeline/pipelines/controller/job_scheduling.py
+++ b/pypeline/pipelines/controller/job_scheduling.py
@@ -12,25 +12,28 @@ from terraform.terraform import Terraform
 
 def main():
     # TODO : Refactor, make function to generate layout
-    azure_eastus2 = Azure(region="eastus2")
+    az_regions = ["eastus2"]
+    az_scenario_name = "job-scheduling"
+    az_deleteion_delay = "1h"
+    azure_cloud = Azure(region=az_regions[0])
     job_scheduling = Benchmark(
         name="job_scheduling",
         layouts=[
             Layout(
                 display_name="azureeastus2",
-                cloud=azure_eastus2,
+                cloud=azure_cloud,
                 setup=Setup(run_id=os.getenv("RUN_ID")),
                 resources=[
                     ResourceGroup(
-                        region="eastus2",
-                        scenario_name="job-scheduling",
-                        deletion_delay="1h",
+                        region=az_regions[0],
+                        scenario_name=az_scenario_name,
+                        deletion_delay=az_deleteion_delay,
                     ),
                     Terraform(
-                        cloud=azure_eastus2,
-                        regions=["eastus2"],
-                        scenario_name="job-scheduling",
-                        deletion_delay="1h",
+                        cloud=azure_cloud,
+                        regions=az_regions,
+                        scenario_name=az_scenario_name,
+                        deletion_delay=az_deleteion_delay,
                     ),
                     Python3(),
                     SSH(cloud="azure"),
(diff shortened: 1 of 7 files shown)


pypeline/terraform/resource_group.py
```diff
@@ -0,0 +1,57 @@
...
+        ).strip(""),
+        condition="ne(variables['SKIP_RESOURCE_MANAGEMENT'], 'true')",
+    )
+
+
+# This is for handling cloud resources at terraform
+@dataclass
+class ResourceGroup(Resource):
+    region: str
+    provider: CloudProvider
```
wonderyl, line 44: Only Azure has the concept of ResourceGroup, AWS doesn't have it, GCP has project, but it's not exactly the same. At this stage, we are safe to assume ResourceGroup only applies to Azure.

pypeline/terraform/resource_group.py
```diff
@@ -0,0 +1,57 @@
...
+
+def delete_resource_group(region: str) -> Script:
+    return Script(
+        display_name="Delete Resource Group",
+        script=dedent(
+            f"""
+            echo "Deleting resources and removing state file before retrying"
+            ids=$(az resource list --location {region} --resource-group $RUN_ID --query [*].id -o tsv)
+            az resource delete --ids $ids --verbose
+            rm -r terraform.tfstate.d/{region}
```
wonderyl, line 33: This should be part of terraform teardown. ResourceGroup and Terraform should be completely decoupled.

pypeline/terraform/resource_group.py
```diff
@@ -0,0 +1,57 @@
...
+        condition="ne(variables['SKIP_RESOURCE_MANAGEMENT'], 'true')",
+    )
+
+
+def delete_resource_group(region: str) -> Script:
+    return Script(
+        display_name="Delete Resource Group",
+        script=dedent(
+            f"""
+            echo "Deleting resources and removing state file before retrying"
```
wonderyl, line 30: I also added a feature to skip delete in https://github.com/Azure/telescope/pull/668/files#diff-43b0a6e7ca97b5e7f9289dc2fc51830da2782e6b82fc5d0d091508c6b60ad70aR39 could you add that as well?

vittoriasalim: I will make a new PR for that

wonderyl: The convention is, you need to add a `// TODO: ` to indicate you are making it in a new PR. Otherwise the reviewer needs to track the comments to find out.

pypeline/pipelines/controller/job_scheduling.py
```diff
@@ -17,13 +18,19 @@ def main():
...
                 setup=Setup(run_id=os.getenv("RUN_ID")),
                 resources=[
+                    ResourceGroup(
+                        region="eastus2",
+                        scenario_name="job-scheduling",
+                        deletion_delay="1h",
+                    ),
                     Terraform(
                         cloud=azure_eastus2,
                         regions=["eastus2"],
```
wonderyl, line 31: You already have region info in Cloud. Regions is duplicated.

pypeline/pipelines/controller/job_scheduling.py
```diff
@@ -17,13 +18,19 @@ def main():
         layouts=[
             Layout(
                 display_name="azureeastus2",
-                cloud=Azure(),
+                cloud=azure_eastus2,
                 setup=Setup(run_id=os.getenv("RUN_ID")),
                 resources=[
+                    ResourceGroup(
+                        region="eastus2",
```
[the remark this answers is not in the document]
vittoriasalim: We agreed to optimize this later (with a function that feed all the input)

pypeline/terraform/terraform.py
```diff
@@ -212,8 +187,9 @@ def generate_apply_or_destroy_script(
     if (
         TerraformCommand.APPLY == command and cloud.provider == CloudProvider.AZURE
     ) or (TerraformCommand.DESTROY == command and cloud.provider == CloudProvider.AWS):
-        error_handling_script = indent(cloud.delete_resource_group(), " " * 16)
 
+        # Todo : replace this with error handling script
+        error_handling_script = indent("", " " * 16)
     return dedent(
         f"""
         set -e
```
wonderyl, line 195: Now we have a problem. deletion_delay and owner are in two places, 1 in job_scheduling.py, one in tfvars file. Ideally we should remove these two variables from tfvars file, but it requires us to set these 2 values via terraform command in line 233. 
It doesn't need to be done in this PR though.

pypeline/terraform/terraform.py
```diff
@@ -246,20 +222,14 @@ def generate_apply_or_destroy_script(
...
-    )
-
-
 # TODO: Add delete_resource_group function and validate_resource_group function
 @dataclass
 class Terraform(Resource):
     cloud: Cloud
     regions: list[str]
     scenario_name: str
+    deletion_delay: str
```
wonderyl, line 231: Is this used in terraform.py now?

pypeline/pipelines/controller/job_scheduling.py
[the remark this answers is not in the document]
wonderyl: You can just pass `azure_eastus2.region` to it.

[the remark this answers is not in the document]
vittoriasalim: Some components just need the first region, I made **az_regions** and passed the regions accordingly. Let me know this change works
image

pypeline/pipelines/controller/job_scheduling.py
vittoriasalim: Some components just need the first region, I made **az_regions** and passed the regions accordingly. Let me know this change works
image

PR #694, opened 2025-06-04
Create a pipeline to list ConfigMaps using Kperf

xinWeiWei24:
This pipeline uses Kperf to list ConfigMaps with the apiserver-cm100 scenario. It is designed for benchmarking by customizing ConfigMap resources, which can affect the response size of listing and help evaluate performance easily.


scenarios/perf-eval/apiserver-cm100/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,35 @@
+scenario_type  = "perf-eval"
+scenario_name  = "apiserver-cm100"
+deletion_delay = "20h"
```
sumanthreddy29, line 3: I see that the test completed in less than an hour. Can you change this deletion delay to 1h, since we don't need this for 20hr.

xinWeiWei24: I noticed the test might take about an hour and a few minutes, so I'll change the duration to 2 hours.

pipelines/perf-eval/API Server Benchmark/apiserver-benchmark-configmaps100.yml
```diff
@@ -0,0 +1,34 @@
+trigger: none
+
+variables:
+  SCENARIO_TYPE: perf-eval
+  SCENARIO_NAME: apiserver-cm100
+
```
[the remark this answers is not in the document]
xinWeiWei24: Add a cron to run every 2 days at 3 AM

sumanthreddy29 (approved):

PR #699, opened 2025-06-06
Add v6 with ubuntu2404

alyssa1303:


modules/terraform/azure/aks-cli/main.tf
```diff
@@ -82,7 +82,7 @@ resource "terraform_data" "enable_aks_cli_preview_extension" {
     EOT
       ) : (
       <<EOT
-      az extension add -n aks-preview --version 14.0.0b2
+      az extension add -n aks-preview --version 18.0.0b10
```
anson627, line 85: do we still need this?  ubuntu24 is public previewed now: https://learn.microsoft.com/en-us/azure/aks/upgrade-os-version?tabs=testing-a-new-os-version%2Croll-back-your-os-version

anson627 (approved):

PR #734, opened 2025-07-02
Add support for AKS auto mode with node pool configurations

rafael-mendes-pereira:
Change AKS CLI module to accept null for default node configuration in order to support AKS automatic mode with default node pools.

This change also introduce new terraform tests to validate the AKS CLI command generation for different scenarios, including overrides and optional parameters.

Note: it adds an input variable `dry_run` to permit to mock the command in the unit tests since its required to test it with terraform apply (not plan)


scenarios/perf-eval/cluster-automatic/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,22 @@
+scenario_type  = "perf-eval"
+scenario_name  = "api_cli_test"
+deletion_delay = "72h"
+owner          = "aks"
+aks_cli_config_list = [
+  {
+    role                          = "client1"
+    aks_name                      = "test"
```
sumanthreddy29, line 8:
```suggestion
    role                          = "client"
    aks_name                      = "aks-automatic"
```
Can you use this names instead of test, so that when we see a cluster we know it's from this test scenario.

scenarios/perf-eval/cluster-automatic/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,22 @@
+scenario_type  = "perf-eval"
+scenario_name  = "api_cli_test"
```
sumanthreddy29, line 2:
```suggestion
scenario_name  = "cluster-automatic"
```
Scenario name doesn't match with folder name

rafael-mendes-pereira: Thanks. That was a leftover from a test

scenarios/perf-eval/cluster-automatic/terraform-inputs/azure.tfvars
```diff
@@ -0,0 +1,22 @@
+scenario_type  = "perf-eval"
+scenario_name  = "cluster-automatic"
+deletion_delay = "72h"
```
sumanthreddy29, line 3: Do we need to keep the cluster for 72hrs? Deletion delay is 72hrs in this case?

sumanthreddy29 (approved):

PR #748, opened 2025-07-18
Add support for OpenCost Live Exporter functionality

rafael-mendes-pereira:
The key changes include the addition of Kubernetes cost analysis collection. This allows to collect cost at different levels (node, pod, container)

### Enhancements to step execution tracking:

* [`steps/calculate-duration.yml`](diffhunk://#diff-c6ff73073b5809245c67a1427322d8a9c2b9187a8444bf6ad1682b3f68e2b752R1-R31): Added a reusable template to calculate the duration of step execution in minutes, using start and end epoch times. The duration is stored in a pipeline variable for subsequent use.
* [`steps/record-start-time.yml`](diffhunk://#diff-b3c0776e68cbfe3eefa9273732962d85469adb99b3d1399fbda9b1643cd69ea0R1-R19): Created a reusable template to record the start time of a step execution as an epoch timestamp, storing it in a pipeline variable.
* [`steps/execute-tests.yml`](diffhunk://#diff-a513485bd29f3a131caa8ee353be121ab7117e57788b699afd1bb2451c551600R16-R29): Integrated the `record-start-time` and `calculate-duration` templates into the test execution workflow to track start time and duration.

### Kubernetes cost analysis improvements:

* [`steps/common/k8s-cost-analysis/collect.yml`](diffhunk://#diff-66cb8211b8e67b6e4d75cd186f59b6e33c75aba8fed13d4708e3595986269112R1-R81): Added a template to deploy the OpenCost service, collect Kubernetes cost data, and store the results in JSON files. Includes port-forwarding and Python script execution for cost analysis.
* [`steps/common/k8s-cost-analysis/open-cost-service.yaml`](diffhunk://#diff-858b2a88e4c028cf8768e51a501b66a5c2f1714cbc40c9dbb167f63747ae9d4eR1-R14): Defined a Kubernetes service manifest for the OpenCost service, specifying a ClusterIP type and port configuration.

### Publishing cost analysis results:

* [`steps/common/k8s-cost-analysis/publish-costs.yml`](diffhunk://#diff-672090cbaed36236ccdf31b0cf90f3dae723cdd750f2b87123b278ad959bb57cR1-R23): Added a template to upload Kubernetes cost analysis results (allocation and assets) to cloud storage, supporting modular cloud configurations.


jobs/competitive-test.yml
```diff
@@ -48,6 +48,9 @@ parameters:
 - name: ssh_key_enabled
   type: boolean
   default: true
+- name: publish_k8s_cost
```
alyssa1303, line 51: In the E2E run, looks like this value wasn't evaluated correctly? Would it be easier if we just handle the logic inside the function directly instead of using template variables?

steps/common/k8s-cost-analysis/collect.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    # Add buffer time for the test duration in minutes (ignore seconds)
+    test_duration_int=${TEST_DURATION_MINUTES%.*}
+    window=$((test_duration_int + 5))
+
+    # Generate output file names with timestamp
+    timestamp=$(date +%s)
+    ALLOCATION_K8S_COST_FILE="${timestamp}_${SCENARIO_NAME}_${RUN_ID}_k8s_cost_allocation.json"
+    ASSETS_K8S_COST_FILE="${timestamp}_${SCENARIO_NAME}_${RUN_ID}_k8s_cost_assets.json"
+
```
alyssa1303, line 40: can we put this part inside python code and just use cluster context to distinguish between cluster as RUN_ID should serve as unique identifier for each run already

jobs/competitive-test.yml
rafael-mendes-pereira: Added it to the thread in Teams

anson627, line 51: consider we don't want to publish cost for majority of test, can we not change job template?

modules/terraform/azure/aks/cost-analysis/open-cost-service.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
anson627, line 1: move this to cost benchmark test case

PR #754, opened 2025-07-22
Add pipeline for A/B testing CNI

alyssa1303:


scenarios/perf-eval/cri-resource-consume/terraform-inputs/aws.tfvars
```diff
@@ -118,5 +118,5 @@ eks_config_list = [{
       name = "kube-proxy"
     }
   ]
-  kubernetes_version = "1.31"
+  kubernetes_version = "1.32"
```
sumanthreddy29, line 121: Did we also update azure to use 1.32?

PR #763, opened 2025-07-25
Add RDMA scenario

alyssa1303:


modules/python/gpu/gpu.py
```diff
@@ -234,6 +285,31 @@ def _create_topology_configmap(
...
+        config_dir: Directory containing custom configuration files
+        version: Version of the EFA device plugin to install
+    """
+    _install_operator(
+        chart_version=version,
+        operator_name="aws-efa-k8s-device-plugin",
+        config_dir=config_dir,
+        namespace="kube-system",
+        repo_name="aws",
+        repo_url="https://aws.github.io/eks-charts",
```
sumanthreddy29, line 303: We can also add this to the constant url file?

alyssa1303: done

PR #780, opened 2025-07-31
Add AKS Store Demo setup

rafael-mendes-pereira:
This pull request introduces a new `AKSStoreDemo` framework for managing Kubernetes deployments and cleanup operations, along with enhancements to the Kubernetes client to support namespace overrides and manifest deletion. Key changes include the addition of a new abstract base class and implementation for single-cluster deployments, as well as updates to the Kubernetes client to improve flexibility and functionality.

### New `AKSStoreDemo` Framework:

* **Introduction of `AKSStoreDemo` base class**: Added an abstract base class in `aks_store_demo.py` to manage Kubernetes deployment and cleanup workflows. It includes methods for setting context, ensuring namespaces, applying manifests, and deleting manifests. Two abstract methods, `deploy` and `cleanup`, are defined for subclass-specific implementations.
* **Implementation of `SingleClusterDemo`**: Created a concrete implementation of `AKSStoreDemo` for single-cluster deployments. It handles manifest application and cleanup in a specific namespace and supports wait conditions for resource readiness.
* **Command-line interface**: Added a `main` function to enable deployment and cleanup actions via command-line arguments, allowing users to specify cluster context, namespace, and action type.

### Enhancements to Kubernetes Client:

* **Namespace override for manifest operations**:
  - Updated `apply_manifest_from_url`, `apply_manifest_from_file`, `delete_manifest_from_url`, and `delete_manifest_from_file` methods to accept an optional `namespace` parameter, enabling overrides of the namespace defined in the manifest. [[1]](diffhunk://#diff-6a9560d11d840ef19d3ad281dc0560c97c1646797822837cf1b50375d0f2cec9L740-R746) [[2]](diffhunk://#diff-6a9560d11d840ef19d3ad281dc0560c97c1646797822837cf1b50375d0f2cec9L820-R891) [[3]](diffhunk://#diff-6a9560d11d840ef19d3ad281dc0560c97c1646797822837cf1b50375d0f2cec9L864-R908)
  - Modified internal helper methods `_apply_single_manifest` and `_delete_single_manifest` to support namespace overrides. [[1]](diffhunk://#diff-6a9560d11d840ef19d3ad281dc0560c97c1646797822837cf1b50375d0f2cec9L1020-R1068) [[2]](diffhunk://#diff-6a9560d11d840ef19d3ad281dc0560c97c1646797822837cf1b50375d0f2cec9L1158-R1209)

* **Manifest deletion from URL**:
  - Added a new `delete_manifest_from_url` method to delete resources defined in a manifest URL, with support for dependency-aware reverse deletion and an optional `ignore_not_found` flag.

These changes provide a robust framework for deploying and managing Kubernetes resources in Azure Kubernetes Service (AKS) environments, while improving the flexibility and usability of the Kubernetes client.


modules/python/aks_store_demo/aks_store_demo.py
```diff
@@ -0,0 +1,255 @@
...
+            logger.info(f"Ensuring namespace '{self.namespace}' exists")
+            execute_with_retries(
+                self.k8s_client.create_namespace,
+                self.namespace
+            )
+        except Exception as e:
+            logger.warning(f"Namespace operation: {e}")
+
+    def apply_manifest(self, manifest_file: str, wait_condition: str = None,
+                      wait_resource: str = None, timeout: str = "300s"):
```
alyssa1303, line 52: why do we have to use this format `300s` instead of just number?

modules/python/aks_store_demo/aks_store_demo.py
```diff
@@ -0,0 +1,255 @@
...
+            {
+                "file": f"{base_path}/aks-store-virtual-worker.yaml",
+                "wait_condition_type": "available", 
+                "wait_resource": "deployment/virtual-worker",
+                "timeout": "120s"
+            },
+            {
+                "file": f"{base_path}/aks-store-virtual-customer.yaml",
+                "wait_condition_type": "available",
+                "wait_resource": "deployment/virtual-customer",
```
alyssa1303, line 130: Why don't just split the resource to match the functionso we don't need to extra split step above?

modules/python/aks_store_demo/aks_store_demo.py
```diff
@@ -0,0 +1,255 @@
...
+    parser.add_argument(
+        "--cluster-context",
+        type=str,
+        default="",
+        help="Kubernetes cluster context to use",
+    )
+    parser.add_argument(
+        "--namespace",
+        type=str,
+        default="default",
```
alyssa1303, line 223: why is this default different from the AbstractClass default? Maybe just set empty value for abstract class to avoid confusion

steps/common/aks-store-demo/single-cluster/configure-demo.yml
alyssa1303, line 1: Why can these 2 files be combined into 1? What is the reason for putting this in common folder instead of putting it under engine folder? I don't see how other tests can utilize this and this isn't really common like a step to validate node count. It's very specific to this demo app

modules/python/aks_store_demo/aks_store_demo.py
rafael-mendes-pereira: This was a leftover created by Claude Sonnet. I'll fix it

modules/python/aks_store_demo/aks_store_demo.py
```diff
@@ -0,0 +1,228 @@
...
+    def deploy(self):
+        """Deploy AKS Store Demo components."""
+
+    @abstractmethod
+    def cleanup(self):
+        """Clean up AKS Store Demo resources."""
+
+
+@dataclass
+class SingleClusterDemo(AKSStoreDemo):
```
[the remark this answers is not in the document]
rafael-mendes-pereira: Idea was to have another scenario deploying the clients of the demo outside the cluster to generate network traffic. That doesn't need to be in another cluster. I'll rename this class to AllInOneAKSStoreDemo to make it clear.

PR #781, opened 2025-07-31
Add GPU module to run nccl-tests

alyssa1303:
This pull request introduces logging improvements, new configuration files, and documentation updates for GPU and network operator setups. Key changes include adding detailed logging for Kubernetes manifest applications, new YAML configurations for GPU and network operators, and a comprehensive README for GPU module usage.

### Logging Improvements:
* Added logging to `_apply_single_manifest` in `kubernetes_client.py` to log the kind, name, and namespace of the manifest being applied.
* Updated the test `test_apply_single_manifest_resource_already_exists` to verify the new logging behavior.

### Documentation Enhancements:
* Added a new `README.md` for the GPU module, providing setup instructions for configuring and executing GPU-related operations.

### GPU Operator Configurations:
* Introduced a `values.yaml` file to configure the GPU operator, including RDMA support and disabling unneeded features like MIG and vGPU.
* Added an MPIJob YAML configuration for running NCCL tests with specific resource and container settings.

### Network Operator Configurations:
* Added a `NicClusterPolicy` YAML file for configuring the network operator, including OFED driver and RDMA shared device plugin settings.
* Introduced a NodeFeatureRule YAML file for labeling nodes with specific PCI device features.
* Updated `values.yaml` for the network operator to enable NFD and disable deployment of NodeFeatureRules.

### DIFF
--- a/modules/python/gpu/gpu.py
+++ b/modules/python/gpu/gpu.py
@@ -176,7 +176,7 @@ def configure(
     config_dir: str,
 ) -> None:
     """
-    Install all required operators with specified versions and custom configurations
+    Install operators with specified versions and custom configurations
 
     Args:
         network_operator_version: Version of the network operator
@@ -184,11 +184,14 @@ def configure(
         mpi_operator_version: Version of the MPI operator
         config_dir: Directory containing custom configuration files
     """
-    install_network_operator(
-        chart_version=network_operator_version, config_dir=config_dir
-    )
-    install_gpu_operator(chart_version=gpu_operator_version, config_dir=config_dir)
-    install_mpi_operator(chart_version=mpi_operator_version)
+    if network_operator_version:
+        install_network_operator(
+            chart_version=network_operator_version, config_dir=config_dir
+        )
+    if gpu_operator_version:
+        install_gpu_operator(chart_version=gpu_operator_version, config_dir=config_dir)
+    if mpi_operator_version:
+        install_mpi_operator(chart_version=mpi_operator_version)
 
 
 def _create_topology_configmap(
@@ -466,19 +469,22 @@ def main():
     configure_parser.add_argument(
         "--network_operator_version",
         type=str,
-        required=True,
+        required=False,
+        default="",
         help="Version of the network operator to install",
     )
     configure_parser.add_argument(
         "--gpu_operator_version",
         type=str,
-        required=True,
+        required=False,
+        default="",
         help="Version of the GPU operator to install",
     )
     configure_parser.add_argument(
         "--mpi_operator_version",
         type=str,
-        required=True,
+        required=False,
+        default="",
         help="Version of the MPI operator to install",
     )
     configure_parser.add_argument(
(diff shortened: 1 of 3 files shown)


modules/python/gpu/gpu.py
```diff
@@ -0,0 +1,562 @@
...
+    """
+    Install all required operators with specified versions and custom configurations
+
+    Args:
+        network_operator_version: Version of the network operator
+        gpu_operator_version: Version of the GPU operator
+        mpi_operator_version: Version of the MPI operator
+        config_dir: Directory containing custom configuration files
+    """
+    install_network_operator(
```
anson627, line 187: make network operator and mpi operator optional?

modules/python/gpu/gpu.py
```diff
@@ -0,0 +1,562 @@
...
+        label_selector="nvidia.com/ofed-driver=", namespace="network-operator"
+    )
+    for pod in mofed_pods:
+        KUBERNETES_CLIENT.run_pod_exec_command(
+            pod_name=pod.metadata.name,
+            namespace="network-operator",
+            command="ibdev2netdev",
+        )
+
+
```
sumanthreddy29, line 76: Can we run these commands during the validate step of the pipeline instead of running it using subprocess here?

anson627 (approved):

anson627:
maybe use python helm sdk instead of helm and subprocess: https://pypi.org/project/pyhelm/

modules/python/gpu/gpu.py
alyssa1303: I can split this install 3 commands maybe so it's easier to choose which to install

alyssa1303:
> maybe use python helm sdk instead of helm and subprocess: https://pypi.org/project/pyhelm/

This module is not from official helm provider, so not safe to use. As discussed, I'll keep this implementation and make sure to capture all output of helm command for debugging. Will also add a version check for helm

anson627 (approved):

modules/python/gpu/config/network-operator/nic-cluster-policy.yaml
```diff
@@ -0,0 +1,49 @@
...
+      autoUpgrade: true
+      maxParallelUpgrades: 1
+      safeLoad: false
+      drain:
+        enable: true
+        force: true
+        timeoutSeconds: 300
+        deleteEmptyDir: true
+  rdmaSharedDevicePlugin:
+    image: k8s-rdma-shared-dev-plugin
```
anson627, line 31: why only shared rdma, but not SR-IOV Device Plugin?

anson627:
I don't see pod-security.kubernetes.io/enforce=privileged on namespace, does that work?

modules/python/gpu/config/network-operator/nic-cluster-policy.yaml
alyssa1303: will add sr-iov in another PR for comparison with rdmad shared device pluging

alyssa1303:
> I don't see pod-security.kubernetes.io/enforce=privileged on namespace, does that work?

We don't need that because our clusters don't use pod security

PR #785, opened 2025-08-01
feat: add containerd metrics to pod slo tests

rbtr:


rbtr:
replacing with in-repo branch instead of fork

PR #787, opened 2025-08-04
Onboard NAP 5k Test Scenario

sumanthreddy29:
This pull request introduces a new performance evaluation pipeline for testing node auto-provisioning (NAP) with a 5000-node Azure Kubernetes Service (AKS) cluster. It includes YAML configurations for scheduling and running the benchmark as well as Terraform input variables for the infrastructure setup. Below are the key changes grouped by their themes:

### Pipeline Configuration
* Added a new YAML pipeline file `node-auto-provisioning-benchmark-5k-nodes.yml` to define two scheduled benchmarks: one for on-demand nodes and another for spot nodes. Each schedule runs daily and includes configurations for the Azure region, cluster loader engine, and node pool settings.

### Infrastructure Setup
* Created a new Terraform variables file `azure-5k.tfvars` to define infrastructure details for the NAP benchmark, such as public IP configurations, virtual network and subnet settings, NAT gateway associations, and AKS cluster configurations. The AKS cluster includes a default node pool and optional parameters for node provisioning and networking.


scenarios/perf-eval/nap/terraform-inputs/azure-5k.tfvars
```diff
@@ -0,0 +1,76 @@
+scenario_type  = "perf-eval"
+scenario_name  = "nap"
+deletion_delay = "4h"
```
alyssa1303, line 3: Is 4h enough? Maybe we should increase to 8h to match the job timeout

sumanthreddy29: Updated 8h

pipelines/perf-eval/Autoscale Benchmark/node-auto-provisioning-benchmark-5k-nodes.yml
```diff
@@ -0,0 +1,99 @@
...
+            - eastus2
+          terraform_input_file_mapping:
+            - eastus2: "scenarios/perf-eval/nap/terraform-inputs/azure-5k.tfvars"
+          engine: clusterloader2
+          engine_input:
+            image: "ghcr.io/azure/clusterloader2:v20250423"
+          topology: karpenter
+          matrix:
+            5k-on-demand:
+              cpu_per_node: 2
```
alyssa1303, line 45: This might have to be variable too if vm_size is not fixed. Although for scheduled runs, I think we should fix these variables here to avoid surprise.

sumanthreddy29: Updated the params to be fixed for scheduled runs.

alyssa1303 (approved):

PR #809, opened 2025-08-12
Update EKS Launch Template

sumanthreddy29:
This pull request enhances the default launch template creation process in the `eks_client.py` module by improving how block device mappings and network interfaces are configured. It also introduces a new helper method to describe EC2 instance types, which is used to dynamically set up network interfaces based on instance capabilities.

**Enhancements to launch template configuration:**

* The `_create_default_launch_template` method now explicitly sets `BlockDeviceMappings` with a `gp3` EBS volume of 1024 GiB, 5000 IOPS, and 200 MB/s throughput, ensuring consistent storage performance for launched instances.
* The method dynamically configures the `NetworkInterfaces` field by querying the maximum number of network interfaces supported by the instance type and attaching that many EFA interfaces, improving network performance and flexibility.

**New helper method:**

* Added the `describe_instance_types` method to retrieve details about one or more EC2 instance types, enabling dynamic configuration based on instance capabilities.


alyssa1303 (commented):
Which part of the code will ensure the gpu node land on private subnet

PR #879, opened 2025-09-26
Improve deployment workload support in CRUD benchmarking framework

diamondpowell:
## Summary

Adds deployment workload support to the CRUD benchmarking framework, enabling Telescope to measure K8s Deployment create/verify latency on AKS node pools. Also fixes a race condition in Azure LRO handling for scale operations.

Branch cleanup note: Rebased and squashed from 141 commits → 4 logical commits for reviewability. All prior review feedback has been addressed.

## Changes

**'modules/python/crud/azure/node_pool_crud.py'**

- Add `create_deployment()` — creates N deployments from a YAML template, validates readiness via `wait_for_condition`, and handles partial failures gracefully
- Derive `[label_selector]` from parameters (no hardcoding)
- Use `[self.step_timeout]` instead of hardcoded timeout
- Use `%-style` logging for consistency with existing methods
- Make `namespace` configurable with `"default"` as default value

**'modules/python/crud/main.py'**
- Add `handle_workload_operations()` dispatcher with `deployment` subcommand
- CLI args: `--number-of-deployments`, ``--replicas``, `--manifest-dir`, `[--node-pool-name]`
- Remove unnecessary `[--deployment-name]` argument
- Add `else` clause returning error on unknown workload commands

**'modules/python/crud/workload_templates/deployment.yml'**
- Parameterized deployment manifest with configurable replicas, label_selector, and namespace

`[steps/engine/crud/k8s/execute.yml]`
- Add deployment script block between scale-up and scale-down operations
- Wire `NUMBER_OF_DEPLOYMENTS`, `REPLICAS`, `MANIFEST_DIR` environment variables

 `[steps/topology/k8s-crud-gpu/execute-crud.yml]`
- Pass deployment parameters from pipeline matrix through topology layer

**'modules/python/clients/aks_client.py'**
- Fix: chain `.result()` on `begin_create_or_update()` in `scale_node_pool` and `_progressive_scale` — blocks until ARM completes before proceeding. Matches existing pattern in `create_node_pool` (line 331) and `delete_node_pool` (line 573) which already call `.result()`. Prevents race condition where `wait_for_nodes_ready` runs before ARM has processed the scale request. Consistent with the internal repo which also calls `.result()` on all operations.

## Tests

Unit tests in `test_azure_node_pool_crud.py` and `test_main.py`:

`test_create_deployment_success` — all deployments ready
`test_create_deployment_failure` — all fail to become ready
`test_create_deployment_partial_success` — continues on individual failures, returns False
`test_multiple_deployment`s — verifies N deployments created sequentially
`test_handle_workload_operations` — deployment command routing
`test_handle_workload_operations_unknown_command` — returns error
`test_progressive_scaling_failure` — scale continues after step failure
`test_scale_down_fails_continues` — delete still runs after scale-down error
`test_returns_false_early_exit` — unknown operation returns False


modules/python/crud/main.py
```diff
@@ -320,6 +344,32 @@ def main():
     )
     all_parser.set_defaults(func=handle_node_pool_operation)
 
+    # Deployment command - add after the "all" command parser
+    deployment_parser = subparsers.add_parser(
+        "deployment", parents=[common_parser], help="create deployments"
+    )
+    deployment_parser.add_argument("--node-pool-name", required=True, help="Node pool name")
+    deployment_parser.add_argument("--deployment-name", required=True, help="Deployment name")
```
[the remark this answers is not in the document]
diamondpowell: its not mentioned in deploy_kwargs. I'll remove it

modules/python/crud/main.py
```diff
@@ -146,6 +146,30 @@ def handle_node_pool_operation(node_pool_crud, args):
...
+
+    try:
+        if command == "deployment":
+            # Prepare deploy arguments
+            deploy_kwargs = {
+                "node_pool_name": args.node_pool_name,
+                "replicas": args.replicas,
+                "manifest_dir": args.manifest_dir,
+                "number_of_deployments": args.number_of_deployments
+            }
```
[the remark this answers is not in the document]
engineeredcurlz: have added this to stop false successes and return logged errors

steps/engine/crud/k8s/execute.yml
```diff
@@ -54,6 +59,35 @@ steps:
...
+    set -eo pipefail
+
+    # Deploy Workloads
+    PYTHONPATH=$PYTHONPATH:$(pwd) python3 "$PYTHON_SCRIPT_FILE" deployment \
+      --cloud "$CLOUD" \
+      --run-id "$RUN_ID" \
+      --result-dir "$RESULT_DIR" \
+      --node-pool-name "$POOL_NAME" \
+      --deployment-name "$DEPLOYMENT_NAME" \
+      --number_of_deployments "$NUMBER_OF_DEPLOYMENTS" \
```
lokesh-keyan: number_of_deployments to number-of-deployments

diamondpowell: Updated to use number-of-deployments and match argparse

diamondpowell:

diamondpowell:
## Addressed Copilot review feedback:

- `--manifest-dir` made optional — defaults to `None`, uses built-in template at `workload_templates/deployment.yml` when not provided
- Deployment step gated on Azure (`${{ if eq(parameters.cloud, 'azure') }}`), `manifest-dir` passed conditionally via `${MANIFEST_DIR:+--manifest-dir "$MANIFEST_DIR"}`
- Per-deployment unique labels (`f"{label_selector.split('=', 1)[-1]}-{deployment_index}"`)
- Template path resolved via `os.path.dirname(__file__)` (no hardcoded paths)
- Namespace passthrough to deployment operations
- Removed empty `name:` field from deployment template spec
- `hasattr(node_pool_crud, 'create_deployment')` guard for backward compatibility
- Renamed test to match actual operation (`test_handle_workload_operations_create_deployment_success`)

### Files changed:
- `modules/python/crud/azure/node_pool_crud.py`
- `modules/python/crud/main.py`
- `modules/python/crud/workload_templates/deployment.yml`
- `modules/python/tests/crud/test_main.py`
- `steps/engine/crud/k8s/execute.yml`

modules/python/crud/main.py
```diff
@@ -320,6 +351,32 @@ def main():
     )
     all_parser.set_defaults(func=handle_node_pool_operation)
 
+    # Deployment command - add after the "all" command parser
+    deployment_parser = subparsers.add_parser(
+        "deployment", parents=[common_parser], help="create deployments"
+    )
+    deployment_parser.add_argument("--node-pool-name", required=True, help="Node pool name")
```
xinWeiWei24: Since the handling function is named handle_workload_operations and is designed to support other operation types in the future, would it be better to name this workload_parser instead of deployment_parser?

steps/topology/k8s-crud-gpu/execute-crud.yml
```diff
@@ -22,3 +22,6 @@ steps:
     result_dir: $(System.DefaultWorkingDirectory)/$(RUN_ID)
     gpu_node_pool: $(GPU_NODE_POOL)
     step_wait_time: $(STEP_WAIT_TIME)
+    number_of_deployments: $(NUMBER_OF_DEPLOYMENTS)
```
xinWeiWei24: Do you need to update it with your custom values in pipelines/perf-eval/GPU Benchmark/k8s-gpu-cluster-crud.yml? I didn’t see any changes here.

modules/python/crud/azure/node_pool_crud.py
```diff
@@ -270,3 +272,127 @@ def all(
...
+
+            # Check if all deployments were successful
+            if successful_deployments == number_of_deployments:
+                logger.info("Successfully created all %d deployment(s)", number_of_deployments)
+                return True
+            if successful_deployments > 0:
+                logger.warning("Created %d/%d deployment(s)", successful_deployments, number_of_deployments)
+                return False
+            logger.error("Failed to create any deployments")
+            return False
```
xinWeiWei24, line 382: This can be merged like created 0/number_of_deployments

modules/python/crud/azure/node_pool_crud.py
```diff
@@ -270,3 +272,127 @@ def all(
...
+
+        except Exception as e:
+            logger.error("Failed to create deployments: %s", e)
+            return False
```
xinWeiWei24: Recommend removing the outer try/except entirely and letting unexpected errors propagate. If we want to keep "one bad deployment doesn't stop the rest" behavior, the inner except Exception is enough

modules/python/crud/azure/node_pool_crud.py
```diff
@@ -270,3 +272,127 @@ def all(
...
+                return False
+
+            successful_deployments = 0
+
+            # Loop through number of deployments
+            for deployment_index in range(1, number_of_deployments + 1):
+                logger.info("Creating deployment %d/%d", deployment_index, number_of_deployments)
+
+                try:
+                    if manifest_dir:
```
xinWeiWei24: better to extract the per-deployment logic into a helper method

modules/python/crud/main.py
diamondpowell: Good suggestion, the docstring already references statefulset and jobs because those are coming in follow-up PRs ([#1132](https://github.com/Azure/telescope/pull/1132) and [#1133](https://github.com/Azure/telescope/pull/1133)). Each workload type has its own parser with unique args. They all route to handle_workload_operations.  Renaming to workload_parser would be ambiguous since there are multiple.

steps/topology/k8s-crud-gpu/execute-crud.yml
diamondpowell: Oh you're right, I didn't set the GPU pipeline matrix setting, which means there were no defaults passing. Added to the azure blocks, skipped aws since deployment is gated to azure

modules/python/crud/azure/node_pool_crud.py
diamondpowell: Good call, merged them. `Created 0/5` gives more info than "Failed to create any" anyway.

modules/python/crud/main.py
xinWeiWei24: Looking at the three parsers side-by-side, almost every argument is duplicated:
--node-pool-name, --manifest-dir, --replicas/--completions (same semantics), and the only real difference is the cosmetic name of --number-of-{deployments,statefulsets,jobs}.

I'd still keep three subparsers (so --help stays clean and future workload-specific flags have a natural home), but suggest extracting a workload_common_parser with --node-pool-name, --manifest-dir, --replicas, --count, --label-selector, and reusing it via parents=[common_parser, workload_common_parser]. Each subparser would then only declare what is genuinely workload-specific.

This addresses the duplication without losing the per-workload --help clarity that motivated keeping them separate.

```python
workload_common_parser = argparse.ArgumentParser(add_help=False)
workload_common_parser.add_argument(
    "--node-pool-name", required=True, help="Node pool name"
)
workload_common_parser.add_argument(
    "--manifest-dir", required=True,
    help="Directory containing Kubernetes manifest files for the workload"
)
workload_common_parser.add_argument(
    "--replicas", type=int, default=10,
    help="Number of replicas per workload (for jobs, maps to spec.completions)"
)
workload_common_parser.add_argument(
    "--count", type=int, default=1,
    help="Number of workloads to create"
)
workload_common_parser.add_argument(
    "--label-selector", default="app=nginx-container",
    help="Label selector for created pods (default: app=nginx-container)"
)

# Deployment subcommand
deployment_parser = subparsers.add_parser(
    "deployment",
    parents=[common_parser, workload_common_parser],
    help="create deployments",
)
```

modules/python/crud/azure/node_pool_crud.py
diamondpowell: Yes this does make more sense. Pulled the actual work (resolve template, build name, apply, wait) into _apply_deployment so now the loop just calls it, counts successes, and catches errors. Also made it raise a TimeoutError when the deployment isn't ready instead of returning False,nthat way the loop's except handles everything in one place.

modules/python/crud/azure/node_pool_crud.py
diamondpowell: Oh yeah good point, the inner except already handles the per-deployment failures so the outer one was just hiding things. Removed it so any setup issues will actually surface instead of silently returning False.

modules/python/crud/main.py
diamondpowell: Ahhhh yeah I see the duplication in each one. Extracted a workload_common_parser with those plus a generic --count for number-of-workloads. Now the deployment parser is just:
```
deployment_parser = subparsers.add_parser(
    "deployment", parents=[common_parser, workload_common_parser], help="create deployments"
)
```
There are now zero unique args. Statefulset and jobs PRs will be the same, just parents=[common_parser, workload_common_parser] and they're done.

Also renamed --number-of-deployments to --count (the other workloads will follow the same) and updated the pipeline YAML chain to match. Made --manifest-dir optional everywhere since all three workloads have built-in templates under workload_templates/.

modules/python/crud/azure/node_pool_crud.py
```diff
@@ -270,3 +272,138 @@ def all(
             logger.error(error_msg)
             errors.append(error_msg)
             return False
+    def _apply_deployment(
```
xinWeiWei24, line 276: Please add a blank line before the function.

modules/python/crud/azure/node_pool_crud.py
```diff
@@ -270,3 +272,138 @@ def all(
...
+        logger.info("Waiting for pods of deployment %s to be ready...", deployment_name)
+        k8s_client.wait_for_pods_ready(
+            operation_timeout_in_minutes=5,
+            namespace=namespace,
+            pod_count=replicas,
+            label_selector=f"app={deployment_label}"
+        )
+
+        logger.info("Successfully created and verified deployment %d", deployment_index)
+    def create_deployment(
```
xinWeiWei24, line 349: Please add a blank line before the function.

modules/python/crud/azure/node_pool_crud.py
diamondpowell: thanks for this catch! added the blank line

modules/python/crud/azure/node_pool_crud.py
diamondpowell: added the blank line here too

xinWeiWei24 (approved):

PR #948, opened 2025-11-26
feat(udr): enable udr+firewall+route table with subnet association

vittoriasalim:
nap-complex test scenarios: enable udr+firewall+route table with subnet association

### DIFF
--- a/scenarios/perf-eval/nap/terraform-inputs/azure-complex.tfvars
+++ b/scenarios/perf-eval/nap/terraform-inputs/azure-complex.tfvars
@@ -47,7 +47,7 @@ network_config_list = [
               {
                 name             = "required-services"
                 source_addresses = ["*"]
-                target_fqdns     = ["*.azure.com", "*.core.windows.net", "*.azurecr.io", "*.ubuntu.com", "AzureKubernetesService"]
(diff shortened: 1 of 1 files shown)


scenarios/perf-eval/nap/terraform-inputs/azure-complex.tfvars
```diff
@@ -12,12 +19,88 @@ network_config_list = [
     subnet = [
```
vittoriasalim: vnet_address_space and nap-subnet-ms , share the same address causing subnet overlap error
```
Virtual Network Name: "nap-vnet-ms"): performing CreateOrUpdate: 
unexpected status 400 (400 Bad Request) with error: SubnetIpAddressesRangeOverlap: 
Subnet /subscriptions/..../nap-subnet-ms is not valid because its IP address range overlaps with that of an existing 
subnet in virtual network /..../nap-vnet-ms.
```

Proposing following adjustment as seen in the right  
VNet: 10.192.0.0/10
 ├─ nap-subnet-ms: 10.192.0.0/16
 └─ AzureFirewallSubnet: 10.193.0.0/26
/10 → 10.192.0.0 – 10.255.255.255
/16 → 10.192.0.0 – 10.192.255.255
/26 → 10.193.0.0 – 10.193.0.63

PR #1034, opened 2026-01-23
[ON HOLD] Modify kubelet with flags using Helm chart and jq during validation phase to improve node pressure management

pdamianov-dev:
This is related to the creation of a suitable framework to run pressure tests against node images as part of Node Hardening efforts. 

Supporting documents: 
- [Hardening-PRD-Telescope-changes-proposal](https://msazure.visualstudio.com/CloudNativeCompute/_wiki/wikis/PersonalPlayground/944847/Hardening-PRD-Telescope-changes-proposal
)
- [Node Hardening PRD](https://microsoft.sharepoint.com/:w:/r/sites/ec5f8c8f-422b-4bd3-9989-ff63e236343f/_layouts/15/doc2.aspx?sourcedoc=%7B04AC643D-A481-4DC6-A322-73B3B9B1104B%7D&file=Node%20Hardening%20PRD.docx&action=default&mobileredirect=true&share=IQE9ZKwEgaTGTaMic7O5sRBLAatcFt_9mEX46sOlFJNX7DI&wdOrigin=TEAMS-MAGLEV.p2p_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1764948118692&web=1)


pipelines/perf-eval/CRI Benchmark/k8s-resource-pressure.yml
```diff
@@ -48,6 +78,8 @@ stages:
               pod_startup_latency_threshold: 23s
               kubernetes_version: "1.34"
               scrape_kubelets: True
+              use_custom_kubelet: ${{ parameters.use_custom_kubelet }}
```
vittoriasalim: this kubelet config is for node image, any reason we are applying at the benchmark I suggest to apply during nodepool creation https://github.com/MicrosoftDocs/azure-aks-docs/blob/main/articles/aks/custom-node-configuration.md#add-a-node-pool-using-custom-configuration-files:~:text=Create%20a%20new%20Linux%20node%20pool%20using%20the%20az%20aks%20nodepool%20add%20command%20and%20specifying%20your%20configuration%20files%20for%20the%20%2D%2Dkubelet%2Dconfig%20and%20%2D%2Dlinux%2Dos%2Dconfig%20parameters.%20The%20following%20example%20command%20creates%20a%20new%20Linux%20node%20pool%20with%20the%20custom%20./linuxkubeletconfig.json%20file%3A

modules/python/clusterloader2/utils.py
```diff
@@ -196,3 +196,95 @@ def parse_xml_to_json(file_path, indent=0):
...
     return json_result
+
+def add_flags_to_daemonset(flags_json_string):
+    kubelet_flags = parse_flags_from_json(flags_json_string)
+    kubelet_daemonset = rf"""
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: kubelet-config-updater
+  namespace: kube-system
```
vittoriasalim: can we move this to validation stage , e.g: https://github.com/Azure/telescope/blob/main/steps/topology/karpenter/validate-resources.yml

vittoriasalim: We have this example where it patch the VM_SIZE to yaml template https://github.com/Azure/telescope/blob/39223da82abc6a7d1516e28ccce9d76378e8c6d0/steps/topology/karpenter/validate-resources.yml#L28

pipelines/perf-eval/CRI Benchmark/k8s-resource-pressure.yml
```diff
@@ -1,8 +1,34 @@
 trigger: none
 
+parameters:
```
vittoriasalim: We do not need parameters , simply enable them in your test

````
            n1-p60-memory-managed:
              node_count: 1
              max_pods: 60
              repeats: 1
              operation_timeout: 3m
              load_type: memory
              pod_startup_latency_threshold: 23s
              kubernetes_version: "1.34"
              k8s_os_disk_type: Managed
              scrape_kubelets: True
              enable_custom_kubelet: false
              kubelet_config_type: "eviction-hard" # eviction-soft, eviction-soft-grace-period
````


in validate-resources 

can call
$KUBELET_CONFIG_TYPE

if KUBELET_CONFIG_TYPE == ëviction-hard : 
  -  override the memory, nodefs and pid 
else if 
else if 
so on and so on

vittoriasalim (changes requested):
remove parameters and only enable those variable inside your test 

rename use_custom_kubelet to enable_custom_kubelet

pipelines/perf-eval/CRI Benchmark/k8s-resource-pressure.yml
pdamianov-dev: What if we want to pass in different flags with each run without changing the pipeline definition?

pdamianov-dev: what if i set a queue time variable instead?

vittoriasalim: if you want to do it manually , i am suggesting to use the test branch. In the main branch, we do not want to override the value during the run (it is hard to maintain) 
We have this pipeline guideline where all additional variables must be stored in the matrix format, we recommend all pipelines to follow this format. If you need to run manually, and modify each values, you can do that with test branch 
Image

vittoriasalim: The variable group are usually for secret key values like credentials and masked value of aks_custom_header. Since you only care about manual run, you can have 3 stages eviction-hard, eviction-soft. eviction-soft-grace-period. And you can select only the one which you want to trigger. 
Image

PR #1035, opened 2026-01-27
Adding unique node subnet for nodepools

shreyashastantram:
Creating Unique subnet for nodepools.
Adding max surge as 1 to extend beyond 3500 nodes for podsubnet mode.


isaac-dasan (commented):
Also please remove the chages to matrix file and new-pipeline-test.yml file before merge

PR #1124, opened 2026-04-08
Remove apiserver-vn10pod100 pipeline and all related files

liyu-ma:
## Summary
- Remove the `apiserver-vn10pod100` (API Server Benchmark, 10 virtual nodes, 100 pods) pipeline, scenario configs, and dedicated Kusto dashboard
- Delete `dashboards/` directory (only contained the vn10pod100 dashboard)
- Update GCP README to reference `apiserver-vn100pod3k` as the example scenario instead


dashboards/example.json
```diff
@@ -1,327 +0,0 @@
-{
-    "$schema": "https://dataexplorer.azure.com/static/d/schema/52/dashboard.json",
-    "id": "5a058a0b-b125-487c-a7c6-13a78f48a4f1",
```
xinWeiWei24, line 3: This is like an example for dashboard, do we need to delete it or we can just change to another scenario

xinWeiWei24 (approved):

dashboards/example.json
liyu-ma: As discussed this does not seem to be used by any teams, remove it.

PR #1127, opened 2026-04-14
Add API server executing seats testing to Telescope V2

xinWeiWei24:


kcl/ccp_team/hyperscale_apiserver_executing_seats/cl2.yaml
```diff
@@ -0,0 +1,100 @@
+apiVersion: v1
```
liyu-ma, line 1: Should move everything to under `/ccp_team` folder

kcl/hyperscale_apiserver_executing_seats/pipeline.k
```diff
@@ -0,0 +1,188 @@
...
+          "managedOutboundIPProfile": {
+            "count": 10
+          }
+        }
+      }
+    }
+  }""")
+
+createClusterScript = """
+az cloud update --endpoint-resource-manager https://eastus2euap.management.azure.com/
```
liyu-ma: ARM endpoint is not required any more.

kcl/lib/steps/k8s/run_cluster_loader_2.k
```diff
@@ -1,13 +1,23 @@
...
+    preApply = "kubectl apply -f $(Pipeline.Workspace)/s/${preManifest}\n" if preManifest else ""
     script = """
+set +eo pipefail
 kubectl create namespace "${namespace}" || true
-kubectl apply -f $(Pipeline.Workspace)/s/${manifest}
+${preApply}kubectl apply -f $(Pipeline.Workspace)/s/${manifest}
 # Wait
 while true; do
-  phases="$(kubectl get pods --namespace="${namespace}" -o jsonpath='{range .items[*]}{.status.phase}{"\\n"}{end}')"
+  phases=""
```
liyu-ma: I don't think the inner retry here provide much benefits. The key here is `set +eo` which is what we need, and there is already a while loop to keep retrying pod status. If you are concerned silent failure with `kubectl get pods` we can do extra logging like 
```
phases="$(kubectl get pods ... 2>&1)" || echo "kubectl get pods failed: $phases"
```
but 5-attempts retry feels unnecessary.

kcl/hyperscale_apiserver_executing_seats/pipeline.k
```diff
@@ -0,0 +1,188 @@
...
+                    "kwok=true"),
+                azure.WaitForNodePoolSucceeded(
+                    const.DEFAULT_SERVICE_CONNECTION,
+                    CLUSTER,
+                    RESOURCE_GROUP,
+                    SUBSCRIPTION_ID,
+                    KWOK_POOL),
+                k8s.CreateKwokNodes(
+                  NODE_COUNT,
+                  logLevel = "$\{{ parameters.log_level }}",
```
liyu-ma: It does not feel necessary to pass in this from pipeline variable. I don't see any downside even we just hardcode it for the whole pipeline.

kcl/lib/steps/k8s/kwok.k
```diff
@@ -2,7 +2,7 @@ import azure_pipelines.ap.steps
 
 KWOK_SCRIPT = "kwok/kwok.py"
 
-CreateKwokNodes = lambda nodeCount: int, params: {str:str} = {} -> steps.Step {
+CreateKwokNodes = lambda nodeCount: int, logLevel: str = "INFO", continueOnError: bool = False, params: {str:str} = {} -> steps.Step {
```
liyu-ma: When would users decide to continue on error, is it when they pass in `validate-success-threshold` explicitly? If yes we don't need to pass in `continueOnError`, because we can just check if param is provided and set value at line 22

kcl/hyperscale_apiserver_executing_seats/pipeline.k
```diff
@@ -0,0 +1,188 @@
...
+    ]
+
+    jobs = [
+        job.Job {
+            job = "benchmarking"
+            displayName = "Benchmarking"
+            timeoutInMinutes = 600
+
+            steps = [
+                common.SetRunId(),
```
liyu-ma, line 75: Lei mentioned to my PR that run_id is not required for v2, and I think that's correct. Run id is only used for resource group, where you can directly assign the default value `$(Build.BuildId)-\${job_id:0:8}` to resource group

kcl/hyperscale_apiserver_executing_seats/pipeline.k
```diff
@@ -0,0 +1,188 @@
+import azure_pipelines.ap
+import azure_pipelines.ap.jobs.job
+import lib.const
+import lib.steps.azure
+import lib.steps.common
+import lib.steps.k8s
+import lib.util
+
+SUBSCRIPTION_ID = "b8ceb4e5-f05b-4562-a9f5-14acb1f24219"
+RESOURCE_GROUP = "$(RUN_ID)"
```
liyu-ma, line 10: As mentioned in another comment, we can directly assign default value to `RESOURCE_GROUP`

kcl/hyperscale_apiserver_executing_seats/pipeline.k
xinWeiWei24: OK! will delete

kcl/lib/steps/k8s/kwok.k
xinWeiWei24: no I think it's still for large scale testing, this is for pipeline and validate-success-threshold is for validation part. There is still some temporary error during creating nodes.

kcl/ccp_team/hyperscale_apiserver_executing_seats/pipeline.k
```diff
@@ -0,0 +1,173 @@
...
+NODE_COUNT = 15000
+KWOK_NODES_PER_CONTROLLER = 100
+CL2_NAMESPACE = "clusterloader2"
+CL2_MANIFEST = "kcl/ccp_team/hyperscale_apiserver_executing_seats/cl2.yaml"
+CL2_TESTCONFIG_MANIFEST = "kcl/ccp_team/hyperscale_apiserver_executing_seats/cl2-testconfig.yaml"
+CL2_POOL = "cl2pool"
+CL2_POOL_VM_SIZE = "Standard_D16S_v4"
+CL2_POOL_COUNT = 10
+KWOK_POOL = "kwokpool"
+KWOK_POOL_VM_SIZE = "Standard_D8_v3"
```
wonderyl: There are variables that are only used once, then it's unnecessary to make a constant. Not only that, it makes reading the code much harder. The reader needs to jump from the pipeline to the definition of the constant, without the help of an IDE.

kcl/ccp_team/hyperscale_apiserver_executing_seats/pipeline.k
```diff
@@ -0,0 +1,173 @@
+import azure_pipelines.ap
+import azure_pipelines.ap.jobs.job
+import lib.const
+import lib.steps.azure
+import lib.steps.common
+import lib.steps.k8s
+import lib.util
+
+SUBSCRIPTION_ID = "b8ceb4e5-f05b-4562-a9f5-14acb1f24219"
+RESOURCE_GROUP = r"$(Build.BuildId)-${job_id:0:8}"
```
wonderyl: You can use run id instead, there's a step for it.

kcl/ccp_team/hyperscale_apiserver_executing_seats/cl2.yaml
```diff
@@ -0,0 +1,100 @@
...
+        resources:
+          requests:
+            cpu: '6'
+            memory: '24Gi'
+        volumeMounts:
+        - mountPath: /override
+          name: cl2-override
+        # Mount a custom config.yaml to override the default CL2 load test config.
+        # This version removes the service creation part, as service IPs become
+        # uncontrollable at large scale with high node and pod counts.
```
wonderyl: This comment is more appropriate to be in the cl2-testconfig.yaml file, please reference which config.yaml it copies from as well.

kcl/ccp_team/hyperscale_apiserver_executing_seats/cl2.yaml
```diff
@@ -0,0 +1,100 @@
...
+        - mountPath: /root/perf-tests/clusterloader2/testing/load/config.yaml
```
wonderyl: could you avoid override the existing config? Give it a diff name to avoid confusion.

kcl/ccp_team/hyperscale_apiserver_executing_seats/cl2-testconfig.yaml
```diff
@@ -0,0 +1,453 @@
+apiVersion: v1
```
wonderyl: Could you merge this into cl2.yaml. it's no need to create two manifests.

kcl/lib/steps/k8s/run_cluster_loader_2.k
```diff
@@ -31,7 +33,7 @@ while true; do
     break
   fi
 
-  sleep 30
+  sleep 120
```
wonderyl: why sleep longer? this is a loop

kcl/lib/steps/k8s/run_cluster_loader_2.k
```diff
@@ -1,13 +1,15 @@
 import azure_pipelines.ap.steps
 import lib.steps.azure
 
-RunClusterLoader2 = lambda serviceConnection: str, namespace: str, manifest: str -> steps.Step {
+RunClusterLoader2 = lambda serviceConnection: str, namespace: str, manifest: str, preManifest: str = "" -> steps.Step {
+    preApply = "kubectl apply -f $(Pipeline.Workspace)/s/${preManifest}\n" if preManifest else ""
```
wonderyl: don't need preapply, you can merge into manifest.

modules/python/utils/logger_config.py
```diff
@@ -24,11 +25,13 @@ def setup_logging():
         "%(asctime)s - %(name)s - %(levelname)s - %(message)s",
         datefmt="%Y-%m-%d %H:%M:%S"
     )
+    level = getattr(logging, os.environ.get("LOG_LEVEL", "INFO").upper(), logging.INFO)
```
wonderyl: could you pass it in from the command line variable instead of using an env var. I know it's a smaller change to use env var, but the env var becomes a hidden backdoor, that is not obvious to other users. They can not use -h to get the env var.

kcl/lib/steps/k8s/kwok.k
wonderyl: how do you distinguish the ones that actually failed node validation?

kcl/lib/steps/k8s/run_cluster_loader_2.k
xinWeiWei24: will change back

kcl/lib/steps/k8s/kwok.k
xinWeiWei24: Split into two parts: allow some errors during creation, and fail with an error if validation does not pass.

kcl/ccp_team/hyperscale_apiserver_executing_seats/pipeline.k
```diff
@@ -0,0 +1,160 @@
...
+az rest \\
+  --method put \\
+  --uri "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GROUP}/providers/Microsoft.ContainerService/managedClusters/${CLUSTER}?api-version=2026-01-02-preview" \\
+  --body "${requestBody}"
+"""
+
+output = ap.Pipeline {
+    name = "Hyperscale API Server Executing Seats"
+
+    trigger =  ["v2"]
```
wonderyl: It's probably too expensive to trigger on every v2 commits. You can remove it, then it's manual trigger

liyu-ma (approved):

wonderyl (approved):

PR #1143, opened 2026-04-22
kwok: fetch controller deployment from source instead of apiserver

xinWeiWei24:
Refactor to avoid mutating a deployment read from kube-apiserver.

### DIFF
--- a/kcl/example_pipeline/pipeline.yaml
+++ b/kcl/example_pipeline/pipeline.yaml
@@ -1,5 +1,4 @@
 name: Example Pipeline
-pool: AKS-Telescope-Ubuntu-EastUS2
 trigger:
 - v2
 parameters:
--- a/kcl/example_pipeline/pipeline.yaml
+++ b/kcl/example_pipeline/pipeline.yaml
@@ -1,4 +1,5 @@
 name: Example Pipeline
+pool: AKS-Telescope-Ubuntu-EastUS2
 trigger:
 - v2
 parameters:
--- a/modules/python/kwok/kwok.py
+++ b/modules/python/kwok/kwok.py
@@ -5,6 +5,8 @@ from abc import ABC, abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass
 
+import json
+
 import requests
 import yaml
 from kubernetes import client
@@ -89,7 +91,7 @@ class KWOK(ABC):
             raise RuntimeError("Controller Deployment not found in kwok.yaml.")
 
         controller_deployment = client.ApiClient().deserialize(
-            type("Resp", (), {"data": yaml.dump(controller_manifest)})(),
+            type("Resp", (), {"data": json.dumps(controller_manifest)})(),
             "V1Deployment",
         )
 
--- a/modules/python/kwok/kwok.py
+++ b/modules/python/kwok/kwok.py
@@ -70,11 +70,26 @@ class KWOK(ABC):
         stage_fast_yaml_url = (f"https://github.com/{self.kwok_repo}/releases/"
                               f"download/{kwok_release}/stage-fast.yaml")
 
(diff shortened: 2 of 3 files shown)


modules/python/kwok/kwok.py
```diff
@@ -60,17 +60,39 @@ def fetch_latest_release(self):
...
+                execute_with_retries(
+                    self.k8s_client.apply_manifest_from_file,
+                    manifest_dict=doc,
+                )
+
+        if controller_manifest is None:
+            raise RuntimeError("Controller Deployment not found in kwok.yaml.")
+
+        controller_deployment = client.ApiClient().deserialize(
+            type("Resp", (), {"data": yaml.dump(controller_manifest)})(),
```
liyu-ma: I think this creates some yaml content, not json? Is it possible to add some unit test to verify this deserialization  is working as expected

modules/python/tests/kwok/test_kwok_node.py
```diff
@@ -448,6 +449,29 @@ def test_tear_down_nodes(self):
...
 
 
+class TestDeploymentDeserialization(unittest.TestCase):
+    """Verify that a YAML-parsed deployment dict deserializes into V1Deployment."""
+
+    def test_deserialize_controller_manifest(self):
+        manifest = client.ApiClient().sanitize_for_serialization(
+            make_base_controller_deployment()
+        )
+        deployment = client.ApiClient().deserialize(
```
wonderyl: Is this testing a function you implement or testing a third party code?

modules/python/tests/kwok/test_kwok_node.py
liyu-ma: Can you check if all the tests in this file passed? I think this mock should be updated since we use http get to fetch kwok.yaml instead

liyu-ma: Same as here.

modules/python/tests/kwok/test_kwok_node.py
xinWeiWei24: Liyu suggested to add a unit test to validate the specific pattern we use in _bootstrap_kwok: parsing YAML into a dict, then serializing to JSON for ApiClient().deserialize

modules/python/kwok/kwok.py
xinWeiWei24: Changed to json.dump and added unit test

modules/python/tests/kwok/test_kwok_node.py
xinWeiWei24: Changed to mock request.get

liyu-ma (approved):

kcl/example_pipeline/pipeline.yaml
```diff
@@ -1,5 +1,4 @@
 name: Example Pipeline
-pool: AKS-Telescope-Ubuntu-EastUS2
```
wonderyl, line 2: why pool is removed?

modules/python/tests/kwok/test_kwok_node.py
wonderyl: Unit test is for testing your own code, not third party code. You can test your function that calls this piece of code instead. in this case `_bootstrap_kwok`

wonderyl: isn't it tested in line 376 already?

kcl/example_pipeline/pipeline.yaml
xinWeiWei24: For testing only — this pool is no longer used and will be reverted before merge.

modules/python/tests/kwok/test_kwok_node.py
xinWeiWei24: Yes, I changed to test in line 360, I will delete TestDeploymentDeserialization part

wonderyl (approved):

PR #1146, opened 2026-04-22
feat: use latest version of nap test

vittoriasalim:
remove k8s version and let az cli pick latest version

### DIFF
--- a/scenarios/perf-eval/nap/terraform-inputs/azure.tfvars
+++ b/scenarios/perf-eval/nap/terraform-inputs/azure.tfvars
@@ -10,7 +10,6 @@ aks_cli_config_list = [
     role               = "nap"
     aks_name           = "nap"
     sku_tier           = "standard"
-    kubernetes_version = "1.33"
     default_node_pool = {
       name       = "system"
       node_count = 5


scenarios/perf-eval/nap/terraform-inputs/azure.tfvars
```diff
@@ -29,7 +29,6 @@ aks_cli_config_list = [
     sku_tier              = "standard"
     subnet_name           = "nap-subnet"
     managed_identity_name = "nap-identity"
-    kubernetes_version    = "1.33"
```
xinWeiWei24, line 32: Did you mean the latest K8S version? the latest version is 1.35

vittoriasalim: remove the version since allen asked for following request "Also, the AKS version is hard coded as 1.33, can we just use something like "latest" to have the test always run with latest AKS version?" 

If version is removed , az cli will always use latest version

xinWeiWei24 (approved):
lgtm

liyu-ma (approved):

PR #1155, opened 2026-04-28
Add partial validation threshold for KWOK nodes

xinWeiWei24:
At large scale, a small number of nodes may fail to become ready due to transient issues. The previous all-or-nothing validation would fail the entire pipeline in such cases. This change adds a configurable success_threshold (0.0–1.0, default 1.0) so operators can allow a percentage of nodes to fail validation without blocking the pipeline. Unit tests are added to cover the new threshold behavior.


modules/python/kwok/kwok.py
```diff
@@ -493,17 +506,28 @@ def _validate_kwok_nodes(self):
...
+        failure_count = 0
+
         for node in kwok_nodes:
             try:
                 self._validate_node_status(node)
                 self._validate_node_resources(node)
                 self._validate_node_schedulable(node)
-            except Exception as e:
-                raise RuntimeError(
-                    f"Validation failed for node {node.metadata.name}: {e}"
```
liyu-ma, line 508: Shall we still log the exception for triaging purpose?

kcl/lib/steps/k8s/kwok.k
```diff
@@ -2,7 +2,7 @@ import azure_pipelines.ap.steps
 
 KWOK_SCRIPT = "kwok/kwok.py"
 
-CreateKwokNodes = lambda nodeCount: int, params: {str:str} = {} -> steps.Step {
+CreateKwokNodes = lambda nodeCount: int, params: {str:str} = {}, validateSuccessThreshold: float = 1.0 -> steps.Step {
```
wonderyl: nit: Generally move var len param last, ins this case validateSuccessThreshold goes before params

wonderyl: You provide the default in `def validate(self, node_validate_success_threshold: float = 1.0):` as well. This may cause confusion when someone change the default and turns out it's not working. So please only keep one.

modules/python/kwok/kwok.py
```diff
@@ -674,6 +698,12 @@ def main():
         default=POD_PLAY_STAGE_PARALLELISM,
         help=f"KWOK podPlayStageParallelism (default: {POD_PLAY_STAGE_PARALLELISM}).",
     )
+    parser.add_argument(
+        "--validate-success-threshold",
+        type=float,
+        default=1.0,
```
wonderyl, line 710: This is the third place where default is provided, keep only one, please.

modules/python/kwok/kwok.py
xinWeiWei24: add it back

modules/python/kwok/kwok.py
xinWeiWei24: Now only set the default value here and didn't change kwok.k

kcl/lib/steps/k8s/kwok.k
xinWeiWei24: Now only set the default value in flags in kwok.py and didn't change kwok.k

xinWeiWei24: Now only set the default value in flags in kwok.py and didn't change kwok.k

wonderyl (approved):

liyu-ma (approved):

PR #1156, opened 2026-04-28
fix: fix karpenter region

vittoriasalim:
fix cores availlability in canada central, there is no 96 cores machine in canadacentral so update to 64coresand fix zonal issue to canadacentral


xinWeiWei24 (approved):
LGTM

PR #1158, opened 2026-04-28
Update condition logic in competitive-test.yml

wolfgang-desalvador:
This pull request updates the `jobs/competitive-test.yml` workflow to improve job cancellation handling and make job conditions more robust. The main changes introduce a configurable cancellation timeout and ensure jobs do not run if already canceled.

Job execution and cancellation improvements:

* Added a new `cancellation_timeout_in_minutes` parameter to allow configuring how long to wait before force-canceling a job, and set its default value.
* Updated the job definition to use the new `cancelTimeoutInMinutes` parameter and changed the job's `condition` to ensure it does not run if already canceled.

### DIFF
--- a/jobs/competitive-test.yml
+++ b/jobs/competitive-test.yml
@@ -36,6 +36,9 @@ parameters:
 - name: timeout_in_minutes
   type: number
   default: 60 # default when not specified is 60 minutes
+- name: cancellation_timeout_in_minutes
+  type: number
+  default: 60 # default when not specified is 5 minutes
 - name: retry_attempt_count
   type: number
   default: 3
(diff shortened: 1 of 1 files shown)


jobs/competitive-test.yml
```diff
@@ -36,6 +36,9 @@ parameters:
 - name: timeout_in_minutes
   type: number
   default: 60 # default when not specified is 60 minutes
+- name: cancellation_timeout_in_minutes
+  type: number
```
vittoriasalim, line 40: why do we need this?

wolfgang-desalvador: As of today, even if we cancel a pipeline, it can stay hanging for a very very long time in case it is in a long running step.

This allows users to be sure it is interrupted in the time they want

vittoriasalim (approved):

PR #1163, opened 2026-05-04
Remove run_id parameter and simplify Set Run ID step

wonderyl:
The run_id parameter was unused in practice. Simplify the step to always auto-generate the run ID from BuildId and JobId.


xinWeiWei24 (approved):
LGTM

PR #1165, opened 2026-05-04
Add --log-level CLI flag to kwok

xinWeiWei24:
Exposes logging verbosity as an explicit `--log-level` argument. `setup_logging()` now accepts an optional `level` parameter (default "INFO").

### DIFF
--- a/modules/python/utils/logger_config.py
+++ b/modules/python/utils/logger_config.py
@@ -15,7 +15,7 @@ class AzureDevOpsFormatter(logging.Formatter):
         # Then add the VSO prefix at the start
         return f"{vso_prefix}{formatted_msg}"
 
-def setup_logging():
+def setup_logging(level="INFO"):
     root_logger = logging.getLogger()
     # Clear any existing handlers
     root_logger.handlers = []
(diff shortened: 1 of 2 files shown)


vittoriasalim (approved):

modules/python/utils/logger_config.py
```diff
@@ -24,11 +24,14 @@ def setup_logging():
         "%(asctime)s - %(name)s - %(levelname)s - %(message)s",
         datefmt="%Y-%m-%d %H:%M:%S"
     )
+    if isinstance(level, str):
+        level = getattr(logging, level.upper(), logging.INFO)
```
wonderyl: Wrong level name will be silently dropped and use info. should throw an exception instead.

modules/python/utils/logger_config.py
```diff
@@ -24,11 +24,14 @@ def setup_logging():
...
+
     handler = logging.StreamHandler()
     handler.setFormatter(formatter)
-    handler.setLevel(logging.INFO)
+    handler.setLevel(level)
```
wonderyl, line 29: https://docs.python.org/3/library/logging.html#logging.Handler.setLevel
> Changed in version 3.2: The level parameter now accepts a string representation of the level such as ‘INFO’ as an alternative to the integer constants such as [INFO](https://docs.python.org/3/library/logging.html#logging.INFO).

wonderyl (approved):

PR #1173, opened 2026-05-08
fix: stop using incompatible aks-preview private build in gpu/ray scheduling scenarios

xuexu6666:
## Summary

- `aks_preview-14.0.0b6` (private build) is incompatible with azure-cli 2.85.0, crashing with `'NoneType' object has no attribute '__dict__'` on argument load before any AKS command runs
- Both `k8s-gpu-scheduling` and `k8s-ray-scheduling` had `use_aks_preview_private_build = true`, pointing to the broken private build
- Switching to `false` uses the public `aks-preview 19.0.0b27`, which is a newer version and supports all required features (`--custom-configuration`, network plugin settings, etc.)

## Test plan

- [ ] Verify `k8s-gpu-scheduling` pipeline runs successfully end-to-end
- [ ] Verify `k8s-ray-scheduling` pipeline runs successfully end-to-end
- [ ] Confirm `az aks create` no longer crashes on argument load

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


xinWeiWei24 (approved):
LGTM

PR #1179, opened 2026-05-14
feat(crud): add AKSMachineClient scaffolding for AKS Machine API

karenychen:
## Summary

First of a planned multi-PR series adding an AKS Machine API client (`AKSMachineClient`) for the CRUD benchmarking framework. This PR delivers the **scaffolding**: the agentpool half of the workflow is fully implemented, and the scale/readiness path is stubbed so it can land in a follow-up PR.

## What's in this PR

**Fully implemented (`modules/python/clients/aks_machine_client.py`):**
- `AKSMachineClient` subclass of `AKSClient` with cached ARM token and a shared `requests.Session` for connection pooling
- `create_machine_agentpool` — PUT a `manualScale=0` agentpool (accepts 200/201/202 to handle ARM async semantics); per-request HTTP timeout capped at `_PER_REQUEST_TIMEOUT_CAP` (60s) so a slow PUT can't consume the polling budget
- `_wait_for_agentpool_provisioning` — poll the agentpool until `Succeeded`, surface `Failed` immediately, tolerate transient absent `provisioningState` until both the consecutive-poll bound and a half-budget wall-clock grace window have elapsed

**Stubbed in this PR (raise `NotImplementedError`, signatures preserved for forward compatibility):**
- `scale_machine`
- `_wait_for_machine_node_readiness`
- `_get_machine_name_prefix`

**Not yet introduced:**
- `_scale_machine_individually`, `_scale_machine_batch`, `_create_single_machine`, `_create_batch_machines`, `_make_batch_request`, `_array_split` — these land with `scale_machine` in PR-2.

## Tests

5 unit tests in `modules/python/tests/test_aks_machine_client.py`:
- `create_machine_agentpool` — success, PUT failure surfaces, provisioning timeout raises
- `scale_machine` raises `NotImplementedError` (guard so `MachineCRUD` wire-up can detect the unimplemented path cleanly)
- `_get_machine_name_prefix` raises `NotImplementedError`

All 5 pass locally; pylint 10.00/10 against the repo `.pylintrc`.

## Follow-up PR

PR-2 (`scale_machine` + batch/individual scale paths + per-machine readiness polling + `_get_machine_name_prefix` implementation) is in flight on a separate branch and will be opened once this one merges.

### DIFF
--- a/modules/python/clients/aks_machine_client.py
+++ b/modules/python/clients/aks_machine_client.py
@@ -180,14 +180,12 @@ class AKSMachineClient(AKSClient):
         ``timeout`` seconds have elapsed since invocation. 4xx GETs are raised
         immediately (e.g. 404 'agentpool not found', 401/403) since retrying
         won't change the outcome; 5xx and other transient failures are logged
-        and retried until the deadline. Missing ``provisioningState`` (which
-        ARM can transiently omit right after PUT acceptance) is tolerated for
-        the first half of ``timeout`` and then treated as Failed; this avoids
-        false negatives on slower control planes without infinite patience.
+        and retried until the deadline. By AKS contract a 200 GET on the
+        agentpool always includes ``properties.provisioningState``; if it is
+        absent on a 200 response, that is a Failed signal and we stop
+        immediately rather than waiting out the timeout.
         """
-        start = time.time()
-        deadline = start + timeout
-        none_grace_deadline = start + timeout / 2
+        deadline = time.time() + timeout
         while time.time() < deadline:
             r = self.make_request("GET", url, timeout=30)
             if r.status_code != 200:
@@ -210,11 +208,10 @@ class AKSMachineClient(AKSClient):
             if state == "Failed":
                 logger.error(f"agentpool provisioning failed: {body}")
                 return False
-            if state is None and time.time() >= none_grace_deadline:
-                logger.warning(
-                    f"agentpool provisioningState absent after "
-                    f"{time.time() - start:.0f}s (past half-budget grace); "
-                    f"treating as Failed."
+            if state is None:
+                logger.error(
+                    f"agentpool 200 GET returned without provisioningState; "
+                    f"treating as Failed. body={body}"
                 )
                 return False
             time.sleep(_POLL_INTERVAL_SECONDS)
--- a/modules/python/clients/aks_machine_client.py
+++ b/modules/python/clients/aks_machine_client.py
@@ -39,7 +39,6 @@ _ARM_BASE = "https://management.azure.com"
 _ARM_SCOPE = "https://management.azure.com/.default"
 _AGENTPOOL_API_VERSION = "2024-06-02-preview"
 _POLL_INTERVAL_SECONDS = 10
-_PROVISIONING_NONE_STATE_LIMIT = 5
 # Per-request HTTP timeout is capped so that a single slow PUT/GET cannot
 # consume the caller's whole ``timeout`` budget before the polling loop starts.
 # The remaining budget is reserved for ``_wait_for_agentpool_provisioning``.
@@ -111,10 +110,10 @@ class AKSMachineClient(AKSClient):
         vm_size: str,
         cluster_name: Optional[str] = None,
         timeout: int = 300,
-    ) -> bool:
+    ) -> None:
         """Convert an existing agentpool to Machines mode via PUT.
 
-        Opens an ``OperationContext`` here, enriches metadata, returns True on
+        Opens an ``OperationContext`` here, enriches metadata, returns on
         success, re-raises on failure (so the context records ``success=False``
         with traceback before the exception propagates to ``MachineCRUD``).
 
@@ -130,9 +129,6 @@ class AKSMachineClient(AKSClient):
                 The per-request HTTP timeout is capped at
                 ``_PER_REQUEST_TIMEOUT_CAP`` so a single slow PUT cannot consume
                 the whole budget before polling starts.
-
-        Returns:
-            True iff the agentpool reaches provisioningState='Succeeded'.
 
         Raises:
             RuntimeError: PUT non-2xx, ARM-reported Failed state, or timeout.
@@ -172,7 +168,6 @@ class AKSMachineClient(AKSClient):
                 op.add_metadata("agentpool_info", self.get_node_pool(
                     agentpool_name, cluster_name).as_dict())
                 op.add_metadata("cluster_info", self.get_cluster_data(cluster_name))
-                return True
             except Exception as e:
                 logger.error(f"Failed to create machine agentpool {agentpool_name}: {e}")
                 raise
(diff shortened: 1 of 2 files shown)


modules/python/clients/aks_machine_client.py
```diff
@@ -0,0 +1,256 @@
...
+        ) as op:
+            try:
+                sub = self.subscription_id
+                url = (
+                    f"{_ARM_BASE}/subscriptions/{sub}/resourceGroups/{self.resource_group}"
+                    f"/providers/Microsoft.ContainerService/managedClusters/{cluster_name}"
+                    f"/agentPools/{agentpool_name}?api-version={_AGENTPOOL_API_VERSION}"
+                )
+                body = {"properties": {"mode": "Machines"}}
+                resp = self.make_request("PUT", url, data=body, timeout=timeout)
```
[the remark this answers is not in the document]
karenychen: Addressed in 05810a3b: capped the per-request PUT timeout at `_PER_REQUEST_TIMEOUT_CAP` (60s) via `min(timeout, _PER_REQUEST_TIMEOUT_CAP)`, and the polling deadline keeps the full `timeout` budget. Doc strings updated to reflect the split.

modules/python/clients/aks_machine_client.py
```diff
@@ -0,0 +1,256 @@
...
+                logger.error(f"agentpool provisioning failed: {body}")
+                return False
+            if state is None:
+                none_count += 1
+                if none_count >= _PROVISIONING_NONE_STATE_LIMIT:
+                    logger.warning(
+                        f"agentpool provisioningState absent for {none_count} consecutive polls; "
+                        f"treating as Failed."
+                    )
+                    return False
```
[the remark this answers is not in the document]
karenychen: Addressed in 05810a3b: the "absent `provisioningState`, treat as Failed" branch now requires BOTH `_PROVISIONING_NONE_STATE_LIMIT` consecutive polls AND `time.time() >= start + timeout/2`. The fixed-count threshold alone no longer trips on slower control planes.

modules/python/clients/aks_machine_client.py
```diff
@@ -0,0 +1,256 @@
...
+            self._token_cache_expiry = float(tok.expires_on)
+            return self._token_cache_value
+
+    def make_request(
+        self,
+        method: str,
+        url: str,
+        data: Optional[Dict[str, Any]] = None,
+        timeout: int = 30,
+    ) -> requests.Response:
```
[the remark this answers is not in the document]
karenychen: Addressed in 05810a3b: switched `make_request` to use `self._session.request(...)` against a per-instance `requests.Session()` held on the client, so TCP/TLS connections pool across polls and the planned parallel scale path in PR-2.

modules/python/clients/aks_machine_client.py
```diff
@@ -0,0 +1,256 @@
...
+        """Scale a Machine-mode agentpool by creating ``scale_machine_count`` machines.
+
+        Stubbed in this PR; implementation lands in a follow-up.
+        """
+        raise NotImplementedError(
+            "scale_machine lands in a follow-up PR"
+        )
+
+    @staticmethod
+    def _get_machine_name_prefix(scale_machine_count: int) -> str:
```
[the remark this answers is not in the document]
karenychen: No longer applicable: `_get_machine_name_prefix` is now stubbed to raise `NotImplementedError` (lands with the scale path in PR-2). The "deterministic 12-char prefix" wording has also been removed from the PR description. Resolving.

modules/python/clients/aks_machine_client.py
```diff
@@ -0,0 +1,282 @@
...
+        least half of ``timeout`` have elapsed; only then is it treated as
+        Failed. This avoids false negatives on slower control planes.
+        """
+        start = time.time()
+        deadline = start + timeout
+        none_grace_deadline = start + timeout / 2
+        none_count = 0
+        while time.time() < deadline:
+            r = self.make_request("GET", url, timeout=30)
+            if r.status_code != 200:
```
liyu-ma, line 191: There is no point retrying if we get 4xx client side errors.

modules/python/clients/aks_machine_client.py
```diff
@@ -0,0 +1,282 @@
...
+        are logged and retried until the deadline. Missing ``provisioningState``
+        (which ARM can transiently omit right after PUT acceptance) is tolerated
+        until BOTH ``_PROVISIONING_NONE_STATE_LIMIT`` consecutive polls AND at
+        least half of ``timeout`` have elapsed; only then is it treated as
+        Failed. This avoids false negatives on slower control planes.
+        """
+        start = time.time()
+        deadline = start + timeout
+        none_grace_deadline = start + timeout / 2
+        none_count = 0
```
liyu-ma: We are doing a none_count check + none_grace_deadline elapsed time check, which is complicated and feels repetitive efforts. Can we just streamline to one approach?

modules/python/clients/aks_machine_client.py
```diff
@@ -0,0 +1,282 @@
...
+        )
+
+    # ---- Machine API: agent pool provisioning ----
+    def create_machine_agentpool(
+        self,
+        agentpool_name: str,
+        vm_size: str,
+        cluster_name: Optional[str] = None,
+        timeout: int = 300,
+    ) -> bool:
```
liyu-ma: I don't see any where `False` is returned (only exception raised), then no need for a return value.

modules/python/clients/aks_machine_client.py
karenychen: Addressed in 1f382eeb: 4xx GETs (404 not-found, 401/403) now raise `RuntimeError` immediately; only 5xx and other transient failures retry until the deadline.

modules/python/clients/aks_machine_client.py
karenychen: Addressed in 1f382eeb: streamlined to a single half-budget wall-clock gate. Dropped `none_count` and `_PROVISIONING_NONE_STATE_LIMIT`; the absent-`provisioningState` branch now triggers only once `time.time() >= start + timeout/2`. This keeps the spirit of tolerating a brief post-PUT empty window without the double-check.

modules/python/clients/aks_machine_client.py
karenychen: Addressed in 1f382eeb: changed return type to `-> None`, dropped the `return True`, and updated the docstring (removed the `Returns:` section). Test updated to drop the `assertTrue` on the discarded return value.

modules/python/clients/aks_machine_client.py
karenychen: Follow-up in 75ffefb6: per AKS design contract, a 200 GET on an agentpool always carries `properties.provisioningState`, so the half-budget grace window was redundant. The absent-`provisioningState` branch on a 200 GET now returns Failed immediately (no waiting). 4xx is still raised immediately; 5xx/JSON-parse still retry until the deadline.

liyu-ma (approved):

PR #1182, opened 2026-05-18
Enable ACL resource consume scenario

realsdx:
Enable ACL resource consume scenario for new ACL distro

This PR also bumps the pinned aks-preview extension version, for newer ACL ossku tag


modules/terraform/azure/aks-cli/main.tf
```diff
@@ -281,7 +281,7 @@ resource "terraform_data" "enable_aks_cli_preview_extension" {
     EOT
       ) : (
       <<EOT
-      az extension add -n aks-preview --version 19.0.0b27
+      az extension add -n aks-preview --version 21.0.0b1
```
vittoriasalim, line 284: is this latest one ? are there ways to make this more extensible. Like add function which always get the latest version

vittoriasalim: left to do and can be done in another PR

realsdx: Although I won't recommend it, on a pipeline running always latest might break the pipeline if there are any breaking changes introduced with the update

vittoriasalim: I mean add a functionality ; if there is no version that is state in end user; then get the latest

modules/terraform/azure/aks-cli/main.tf
```diff
@@ -281,7 +281,7 @@ resource "terraform_data" "enable_aks_cli_preview_extension" {
...
       az version
     EOT
```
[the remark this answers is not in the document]
realsdx: Done

PR #1183, opened 2026-05-18
[feat] add utilities on worker pool creation etc

liyu-ma:
- Update `CreateNodePool` to allow `--no-wait`
- Add `PlanNodePools` function to decide pool config based on max node per pool
- Add `WaitForAllWorkerNodesReady` to wait for pools + nodes to be ready
- Updated existing pipelines with usage of `CreateNodePool`


kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
```diff
@@ -10,18 +14,59 @@ while true; do
...
+echo "Waiting for ${expectedCount} nodes in pool ${poolName} to be Ready..."
+while true; do
+  READY=$(kubectl get nodes --no-headers 2>/dev/null \\
+    | awk '$1 ~ /^aks-${poolName}-/ && $2 == "Ready" {c++} END {print c+0}')
+  NOT_READY=$(( ${expectedCount} - READY ))
+  if [ "$NOT_READY" -le 0 ]; then
+    echo "Pool ${poolName}: all ${expectedCount} nodes are Ready."
+    break
+  fi
+  echo "Pool ${poolName}: $READY / ${expectedCount} Ready ($NOT_READY not ready). Retry in 30 seconds"
```
xinWeiWei24: What if there's an infinite loop—do we intend to rely on a pipeline timeout to stop it?

kcl/lib/steps/common/record_benchmark_start.k
```diff
@@ -0,0 +1,16 @@
+import azure_pipelines.ap.steps
+
+# RecordBenchmarkStart emits a step that captures the current UTC time as
+# the BENCHMARK_START pipeline variable, for use downstream (e.g. as the
+# start of a dashboard time range or metrics query window).
+RecordBenchmarkStart = lambda -> steps.Step {
+    steps.Bash {
```
xinWeiWei24: Do we expect more uses in the future — e.g. benchmark start/end, deploy start/end, etc. — would it make sense to abstract this into a more generic helper that just records the current time into a pipeline variable,

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
liyu-ma: I would say yes. This is only used within `WaitForAllWorkerNodesReady` which means user decides to wait for _everything_ to be ready. They can choose to just use `WaitForNodePoolSucceeded` which does not wait for nodes.

kcl/lib/steps/common/record_benchmark_start.k
liyu-ma: Good point. I updated it to record current time. Thanks!

kcl/lib/steps/azure/create_node_pool.k
```diff
@@ -12,8 +22,55 @@ az aks nodepool add \\
...
+  ${flags}
 """
     AzCli(serviceConnection, "Create node pool ${poolName}", script)
 }
+
+# PlanNodePools expands a {sku: totalNodes} map into an ordered list of
+# NodePool values, one per node pool to create. Each SKU is split into
+# ceil(totalNodes / maxNodesPerPool) pools, with the last pool possibly
+# smaller. Pool names are assigned globally as workerpool0, workerpool1, ...
+PlanNodePools = lambda skuToNodes: {str:int}, maxNodesPerPool: int = 800 -> [NodePool] {
```
wonderyl, line 37: This is quite complex, could you write a test for it?
https://www.kcl-lang.io/docs/tools/cli/kcl/test

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
```diff
@@ -10,18 +14,59 @@ while true; do
...
+  sleep 30
+done
+"""
+}
+
+# Wait for each NodePool in pools to reach provisioningState=Succeeded and then have its
+# expected number of nodes Ready via kubectl. The caller must have run
+# GetCredentials earlier in the job so kubectl is configured.
+WaitForAllWorkerNodesReady = lambda serviceConnection: str, cluster: str, resourceGroup: str, subscription: str, pools: [NodePool] -> steps.Step {
```
wonderyl: why not create each waiting as a separate step?

kcl/lib/steps/azure/create_node_pool.k
```diff
@@ -12,8 +22,55 @@ az aks nodepool add \\
...
+        sku = skus[idx]
+        count = counts[idx]
+    } for idx in range(len(skus))]
+}
+
+# CreateNodePools emits one `az aks nodepool add --no-wait` step per NodePool,
+# so all pools provision in parallel. Typical usage is to pass the
+# [NodePool] returned by PlanNodePools, then pair this with
+# WaitForAllWorkerNodesReady (with the same [NodePool]) to gate on completion.
+CreateNodePools = lambda serviceConnection: str, cluster: str, resourceGroup: str, subscription: str, pools: [NodePool], taintPrefix: str = "", labels: str = "" -> [steps.Step] {
```
wonderyl: Why not change the interface of `CreateNodePool` to take NodePool, then user can write the for loop themselves.

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
```diff
@@ -10,18 +14,59 @@ while true; do
...
+    waitOnePool = lambda p: NodePool -> str {
+        waitForNodePoolScript(cluster, resourceGroup, subscription, p.name) + waitForPoolNodesReadyScript(p.name, p.count)
```
wonderyl: why not having a separate waitForPoolNodeReady step, so that use can choose use it or not, or how to use it?

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
xinWeiWei24: Understood—I’m just checking whether there’s code to handle errors.

kcl/lib/steps/azure/create_node_pool.k
liyu-ma: Added test

kcl/lib/steps/azure/create_node_pool.k
liyu-ma: Took the suggestion now `CreateNodePool` takes `NodePool` param. Also updated existing pipelines.

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
liyu-ma: Ok I removed `WaitForAllWorkerNodesReady` func. Added func `WaitForPoolNodesReady` to wait for a single pool to be `Succeeded` and all nodes ready.

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
liyu-ma: Refactored as suggested. Now removed the "wait for all" function. Users can decide what pool they would want to wait.

kcl/example_pipeline/pipeline.k
```diff
@@ -115,10 +115,7 @@ output = ap.Pipeline {
                     CLUSTER,
                     RESOURCE_GROUP,
                     SUBSCRIPTION_ID,
-                    CL2_POOL,
-                    "Standard_D8S_v4",
-                    4,
-                    CL2_TAINT_PREFIX),
+                    azure.NodePool { name = CL2_POOL, sku = "Standard_D8S_v4", count = 4, taintPrefix = CL2_TAINT_PREFIX }),
```
wonderyl: nit: better format? it's 4 items, one per line would be nice.

kcl/lib/steps/azure/plan_node_pools_test.k
```diff
@@ -0,0 +1,54 @@
...
+    assert [p.sku for p in pools] == ["Standard_D8s_v4", "Standard_D8s_v4", "Standard_D8s_v4"]
+}
+
+test_multiple_skus = lambda {
+    pools = PlanNodePools({"Standard_D8s_v4": 1700, "Standard_D4s_v4": 800})
+    assert len(pools) == 4
+    # workerpool indices are globally assigned across SKUs
+    assert [p.name for p in pools] == ["workerpool0", "workerpool1", "workerpool2", "workerpool3"]
+    # 1700 -> 800+800+100, 800 -> 800
+    assert sum([p.count for p in pools], 0) == 2500
```
wonderyl: The assert is too loose. It's better to test the exact size instead of the sum.

wonderyl: In addition, please test the skus as well

kcl/lib/steps/azure/plan_node_pools_test.k
```diff
@@ -0,0 +1,54 @@
...
+}
+
+test_skips_zero_count_skus = lambda {
```
wonderyl, line 37: please use strict testing criteria in all the test, AI would make it super easy

kcl/lib/steps/azure/plan_node_pools_test.k
liyu-ma: Sure will do!

kcl/lib/steps/azure/plan_node_pools_test.k
liyu-ma: Now updated the tests to assert whole list equality.

kcl/lib/steps/azure/plan_node_pools_test.k
liyu-ma: Updated

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
```diff
@@ -1,7 +1,11 @@
 import azure_pipelines.ap.steps
 
-WaitForNodePoolSucceeded = lambda serviceConnection: str, cluster: str, resourceGroup: str, subscription: str, poolName: str -> steps.Step {
-    script = """
+# waitForNodePoolScript renders a bash snippet that polls a single AKS node
+# pool's provisioningState every 30s, exiting 0 on Succeeded and 1 on
+# Failed/Canceled. The pool name is expanded at KCL render time so the emitted
+# script has no shell parameters.
```
wonderyl: The comment is not necessary, it just repeats what the code is writing. Then when the code change, the comments need to change as well, which creates extra work. A good comment either conveys intent or explain a complex algorithm, which is not obvious in code.

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
```diff
@@ -10,18 +14,49 @@ while true; do
...
+    )
+}
+
+# WaitForPoolNodesReady waits for a single NodePool to reach
+# provisioningState=Succeeded and then have all of its expected nodes Ready
+# via kubectl. The caller must have run GetCredentials earlier in the job so
+# kubectl is configured.
+WaitForPoolNodesReady = lambda serviceConnection: str, cluster: str, resourceGroup: str, subscription: str, pool: NodePool -> steps.Step {
+    script = """
+${waitForNodePoolScript(cluster, resourceGroup, subscription, pool.name)}
```
wonderyl: So it will wait for node pool first, then wait for nodes? can it be done in two separate steps, instead of using script concat?

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
```diff
@@ -10,18 +14,49 @@ while true; do
...
+        "Wait for node pool ${poolName} to succeed",
+        waitForNodePoolScript(cluster, resourceGroup, subscription, poolName),
+    )
+}
+
+# WaitForPoolNodesReady waits for a single NodePool to reach
+# provisioningState=Succeeded and then have all of its expected nodes Ready
+# via kubectl. The caller must have run GetCredentials earlier in the job so
+# kubectl is configured.
+WaitForPoolNodesReady = lambda serviceConnection: str, cluster: str, resourceGroup: str, subscription: str, pool: NodePool -> steps.Step {
```
wonderyl: @xinWeiWei24 had a use case, where she wants to wait for a portion of nodes ready, e.g. 90%, then proceed, since wait for a large amount will take a long time.
How can we make this function to be flexible for that use case as well?

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
liyu-ma: We have func `WaitForNodePoolSucceeded` above if users just want to wait for the pool state. But for nodes readiness we don't want to wait when the pool failed (or from creating -> failed). The `waitForNodePoolScript` takes minimal time if pool is succeeded.

xinWeiWei24 (approved):
LGTM

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
liyu-ma: Added `readyThreshold` param to support perc based waiting

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
liyu-ma: removed

kcl/lib/steps/azure/wait_for_node_pool_succeeded.k
wonderyl: By default a step won't run if the previous step failed.

liyu-ma: As suggested, moved the pool status check out of `WaitForPoolNodesReady`

wonderyl (approved):

PR #1164, opened 2026-05-04
feat: added stage for h4 and h8 pod scheduling benchmark

vittoriasalim:

PR #1151, opened 2026-04-27
Add pipeline for hyperscale etcd storage benchmarking

liyu-ma:
### Context
Create pipeline to automate the etcd storage benchmarking for hyperscale cluster.

### Changes
- Add pipeline.k and yaml file
- Allow ingestion of configmaps, events and secrets data, with parameterized storage size
- Add utility to run kperf with internal batching and retry
- Utility construct to print out ASI managed cluster and apiserver analyzer URLs

PR #1184, opened 2026-05-18
Add API server list configmaps benchmark pipeline to Telescope V2

LeonardCareer:
Migrate the V1 apiserver-benchmark-configmaps100 pipeline (ADO pipeline ID 38) to Telescope V2.

### Changes
* New pipeline at kcl/apiserver_benchmark/configmaps100/pipeline.k
* Added reusable RunKperfBenchmark and CollectKperfResult steps
* Weekly schedule at 3:00 UTC Sunday

### What it does
Benchmarks API Server list configmaps performance under two flow control settings:
* workload-low:1000
* exempt:5

### Validation
[pipeline run](https://dev.azure.com/akstelescope/telescope/_build/results?buildId=67668&view=results) passed on ADO (build #67668)
Results uploaded to telescopev2 storage account and kusto
Data matches V1 output format

### Post-merge TODO
- [ ] Update ADO pipeline parameters (branch, variables) to point to v2


LeonardCareer:

kcl/lib/steps/k8s/runkperf.k
```diff
@@ -78,3 +79,73 @@ echo "Ingest complete: $completed ${dataType}s created."
...
+
+# RunKperfBenchmark runs runkperf bench with the given subcommand and flow control settings.
+# Results are written to /tmp/tmp_results.json for collection by CollectKperfResult.
+RunKperfBenchmark = lambda runnerImage: str, flowcontrol: str, subcmd: str, subcmdArgs: str -> steps.Step {
+    steps.Script {
+        script = """
+set -euo pipefail
+set -x
+
+sudo -E $(command -v runkperf) -v 3 bench \\
```
liyu-ma: do we need `sudo` here

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+            runkperf.CollectKperfResult(
+                const.DEFAULT_SERVICE_CONNECTION,
+                flowcontrol,
+                "--total 10000 --size=1024 --group-size=10 --configmap-amount=100",
+                CLUSTER,
+                RESOURCE_GROUP,
+                SUBSCRIPTION_ID),
+            common.UploadResult(
+                const.DEFAULT_SERVICE_CONNECTION,
+                "telescopev2",
```
liyu-ma: Hmm I think we should add these 3 magic string to this repo's const.k, and make the [internal one](https://msazure.visualstudio.com/CloudNativeCompute/_git/telescope?path=/kcl/lib/const.k&version=GBv2&_a=contents) reference them.

kcl/lib/steps/k8s/runkperf.k
```diff
@@ -78,3 +79,73 @@ echo "Ingest complete: $completed ${dataType}s created."
...
+  --kubeconfig ~/.kube/config \\
```
xinWeiWei24: maybe no need to add this flag, since ~/.kube/config is the default value

kcl/lib/steps/k8s/runkperf.k
```diff
@@ -78,3 +79,73 @@ echo "Ingest complete: $completed ${dataType}s created."
...
+  --runner-image="${runnerImage}" \\
+  --runner-flowcontrol="${flowcontrol}" \\
+  --result /tmp/tmp_results.json \\
+  ${subcmd} ${subcmdArgs}
```
xinWeiWei24: Minor nit: would it be clearer to use names like scenario and scenarioArgs? I almost forgot the runkperf subcommand as well.

kcl/lib/steps/k8s/runkperf.k
```diff
@@ -78,3 +79,73 @@ echo "Ingest complete: $completed ${dataType}s created."
...
+    private_fqdn: $private_fqdn,
+    k8s_version: $k8s_version,
+    sku_tier: $sku_tier,
+    subscription: $subscription,
+    cluster_id: $cluster_id
+  }')
+
+# Check for errors and upload error raw data if error rate is too high
+errors_number=$(jq -r '.result.errors | length' /tmp/tmp_results.json)
+error_rate_too_high=$(jq -r '(1 - .result.total / (.loadSpec.count * .loadSpec.loadProfile.spec.total)) > 0' /tmp/tmp_results.json)
```
xinWeiWei24: How do we define “too high”? Also, should this be calculated as a float or an int?

kcl/lib/steps/k8s/runkperf.k
LeonardCareer: V1 also uses sudo here: https://github.com/Azure/telescope/blob/main/steps/engine/kperf/execute.yml#L82

Without it, runkperf might hit permission denied errors. Want me to try removing it and see if it still passes?

kcl/lib/steps/azure/resource_group.k
```diff
@@ -6,7 +6,7 @@ az group create --name "${name}" --location "${location}" --subscription "${subs
 """
     AzCli(
         serviceConnection,
-        "Create resource group ${name} in ${location} (${subscription})",
+        "Create resource group in ${location}",
```
xinWeiWei24: why do we need to delete this

kcl/lib/steps/k8s/runkperf.k
LeonardCareer: V1 also has it explicitly: https://github.com/Azure/telescope/blob/main/steps/engine/kperf/execute.yml#L83. But yeah, it should be the default, I am happy to remove it if it preferred.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+
+SUBSCRIPTION_ID = "b8ceb4e5-f05b-4562-a9f5-14acb1f24219"
+RESOURCE_GROUP = "$(RUN_ID)"
+LOCATION = "uksouth"
+CLUSTER = "configmaps100"
+RUNNER_IMAGE = "ghcr.io/azure/kperf:0.3.4"
+
+requestBody = util.escapeStr("""
+{
+    "location": "${LOCATION}",
```
xinWeiWei24: Didn’t see a nodepool for the kperf runner. Also, it might be better to use az CLI to create the cluster—no need to enable preview features via AZ REST, and it’s cleaner and more straightforward.

kcl/lib/steps/azure/resource_group.k
LeonardCareer: The name here is "$(RUN_ID)", which is an ADO runtime variable, KCL can't resolve it at compile time, so the displayName would show the literal string $(RUN_ID) instead of the actual value. 

Here are two examples where the displayName shows unresolved variables:

https://dev.azure.com/akstelescope/telescope/_build/results?buildId=67523&view=logs&j=19788ade-6ddf-5515-c2e0-d65408b9c5ef&t=ef68049a-80e9-581f-d98d-29549168fe70
https://dev.azure.com/akstelescope/telescope/_build/results?buildId=67255&view=logs&j=f6142361-75fe-5ddc-eae3-6e880e639a5e&t=3f8c6cdc-20e4-5196-4be2-5e66e7bb65b8
That said, this fix isn't really related to this task. Happy to split it into a separate PR if the team prefers.

kcl/lib/steps/k8s/runkperf.k
LeonardCareer: interesting part, but yes, any lost request means "too high"

This is carried over from V1 (https://github.com/Azure/telescope/blob/main/steps/engine/kperf/collect.yml). It calculates the request loss rate: 1 - actual/expected. "Too high" here means any loss at all (> 0). The jq division is float so no int truncation issue. It's just a gate for whether to dump error details

kcl/lib/steps/k8s/runkperf.k
LeonardCareer: Good suggestion. V1 also uses benchmark_subcmd/benchmark_subcmd_args, but I agree scenario/scenarioArgs is clearer. Will update the name and all related parts.

kcl/apiserver_benchmark/configmaps100/pipeline.k
LeonardCareer: Agreed, good suggestion. Currently only my pipeline and the example pipeline have these hardcoded, but high likely all other v2 pipelines will need these variables to upload results.

[example pipeline](https://github.com/Azure/telescope/blob/v2/kcl/example_pipeline/pipeline.k#L18-L20) is here. Should I update the example pipeline as well in this PR, or keep it separate?

kcl/lib/steps/k8s/runkperf.k
xinWeiWei24: yes we can also use this benchmark_subcmd/benchmark_subcmd_args

kcl/lib/steps/k8s/runkperf.k
xinWeiWei24: I still think we should only enable detailed error logging when it’s “too high”—maybe define a threshold, e.g., 10% error rate.

kcl/lib/steps/azure/resource_group.k
xinWeiWei24: i see. but i found subscription still works, do we need to delete the sub as well?

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+        }
+      ],
+      "networkProfile": {
+        "networkPlugin": "azure",
+        "networkPluginMode": "overlay"
+      }
+    }
+  }""")
+
+createClusterScript = """
```
wonderyl, line 19: The example pipeline was using rest API because we were testing an unreleased feature. You can use `az aks create` directly.

kcl/lib/steps/k8s/runkperf.k
wonderyl: At lease we should understand why sudo is needed. Ideally remove it. If not possible, comment it why it's needed.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+                "runner",
+                "Standard_D16s_v3",
+                3),
+            azure.WaitForNodePoolSucceeded(
+                const.DEFAULT_SERVICE_CONNECTION,
+                CLUSTER,
+                RESOURCE_GROUP,
+                SUBSCRIPTION_ID,
+                "runner"),
+            runkperf.InstallRunkperf(RUNNER_IMAGE),
```
wonderyl: I haven't try this before, but azure pipeline can run a container directly. https://learn.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops&tabs=windows
I wonder if we can leverage this feature.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+import lib.steps.azure
+import lib.steps.common
+import lib.steps.k8s.runkperf
+import lib.util
+
+SUBSCRIPTION_ID = "b8ceb4e5-f05b-4562-a9f5-14acb1f24219"
+RESOURCE_GROUP = "$(RUN_ID)"
+LOCATION = "uksouth"
+CLUSTER = "configmaps100"
+RUNNER_IMAGE = "ghcr.io/azure/kperf:0.3.4"
```
wonderyl: I'm an Agent, trained with yaolei@'s past comments.

Several of these constants are used only once or twice (`LOCATION`, `CLUSTER`, `RUNNER_IMAGE`). As I noted on #1127 and #1151, single-use constants make the pipeline harder to read because the reader has to jump back to the top of the file (often without an IDE) to learn what they mean. Please consider inlining the ones used only once.

Also, `RESOURCE_GROUP = "$(RUN_ID)"` is a constant aliasing a pipeline variable — that indirection is confusing. You already call `common.SetRunId()`; use `"$(RUN_ID)"` directly at the call sites, or wrap it in a helper that knows the RG name is the run id.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+      }
+    }
+  }""")
+
+createClusterScript = """
+az rest \\
+  --method put \\
+  --uri "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GROUP}/providers/Microsoft.ContainerService/managedClusters/${CLUSTER}?api-version=2026-01-02-preview" \\
+  --body "${requestBody}"
+"""
```
wonderyl: I'm an Agent, trained with yaolei@'s past comments.

This ~40-line `requestBody` + raw `az rest` PUT is hard to read and bypasses any reusable AKS creation step. We already have `azure.CreateCluster` / similar helpers in `lib/steps/azure`. If the missing capability is the preview API version or one of these fields (e.g. `tags`, `networkPluginMode`), please extend the lib helper instead of inlining `az rest` here — same pattern I called out on !15370069 (`cluster_helpers.k:255`: "hyperscale cli is released, please use that") and on the resource_group/CreateNodePool reuse threads.

wonderyl (commented):
I'm an Agent, trained with yaolei@'s past comments.

A few high-level points before the inline comments:

- Daily 3AM cron may be too expensive — each fire creates two fresh AKS clusters with a `Standard_D16s_v3 x3` node pool. Consider weekly or manual-trigger only (similar to what we landed on in #1127).
- The two stages could share one cluster sequentially. That would halve provisioning cost and give apples-to-apples control-plane state across both flow-control runs.
- The Post-merge TODO ("Update ADO pipeline parameters") is a footgun — if the v1 pipeline keeps running against v2 branch by accident (or vice versa), we'll double-bill / lose data. Worth landing the ADO param change first or scheduling on a different branch until cutover.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+makeStage = lambda flowcontrol: str, flowcontrolLabel: str -> ap.Stage {
+    ap.Stage {
+        stage = flowcontrolLabel
+        displayName = flowcontrolLabel
+        jobs = [makeBenchmarkJob(flowcontrol, flowcontrolLabel)]
+    }
+}
```
wonderyl, line 109: I'm an Agent, trained with yaolei@'s past comments.

- `-> ap.Stage` annotation is unnecessary, you can drop it (same nit as #1164 `pipeline.k:163`/`:167`).
- Since `flowcontrolLabel` is purely derived from `flowcontrol` (and `stage`/`displayName` are identical), consider deriving the label inside the lambda so the call site only passes `flowcontrol`.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+        displayName = "Daily 3AM UTC"
+        branches = ["v2"]
+        always = True
+    }]
+    pool = const.DEFAULT_POOL
+
+    stages = [
+        makeStage("workload-low:1000", "workload_low"),
+        makeStage("exempt:5", "exempt"),
+    ]
```
wonderyl, line 133: I'm an Agent, trained with yaolei@'s past comments.

Use a for-comprehension instead of two `makeStage` calls — same suggestion as #1164:

```kcl
flowcontrols = [("workload-low:1000", "workload_low"), ("exempt:5", "exempt")]
stages = [makeStage(fc, label) for fc, label in flowcontrols]
```

It's more obviously list-shaped and easier to extend with a third flow-control case.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+output = ap.Pipeline {
+    name = "API Server List Configmaps"
+
+    trigger = "none"
+    schedules = [ap.Cron {
+        cron = "0 3 * * *"
+        displayName = "Daily 3AM UTC"
+        branches = ["v2"]
+        always = True
+    }]
```
wonderyl, line 127: I'm an Agent, trained with yaolei@'s past comments.

Daily 3AM is likely too expensive (each fire creates two fresh AKS clusters with `Standard_D16s_v3 x3` node pools). I made a similar comment on #1127. Consider dropping the cron and keeping it manual-trigger, or moving to weekly until we have a real reason to track daily.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+                flowcontrol,
+                "list_configmaps",
+                "--total 10000 --size=1024 --group-size=10 --configmap-amount=100"),
+            runkperf.CollectKperfResult(
+                const.DEFAULT_SERVICE_CONNECTION,
+                flowcontrol,
+                "--total 10000 --size=1024 --group-size=10 --configmap-amount=100",
+                CLUSTER,
+                RESOURCE_GROUP,
+                SUBSCRIPTION_ID),
```
wonderyl: I'm an Agent, trained with yaolei@'s past comments.

`subcmdArgs` (`--total 10000 --size=1024 --group-size=10 --configmap-amount=100`) is repeated verbatim across `RunKperfBenchmark` and `CollectKperfResult`. Lift it to a local in `makeBenchmarkJob` so the two never drift — same spirit as #1151:46 ("redundant, why not use totalGiB directly?").

kcl/lib/steps/azure/resource_group.k
```diff
@@ -16,7 +16,7 @@ az group delete --name "${name}" --subscription "${subscription}" --yes
 """
     AzCli(
         serviceConnection,
-        "Delete resource group ${name} (${subscription})",
+        "Delete resource group",
```
wonderyl: I'm an Agent, trained with yaolei@'s past comments.

Two concerns about this lib change:

1. Dropping `${name}` / `${subscription}` from the displayName loses signal when scanning long pipeline logs — especially when a job creates multiple RGs or operates across subscriptions. The old name was genuinely useful.
2. As I noted on #1151 (`az_cli.k:14`): after updating the lib, please regenerate yaml for **all** pipelines so reviewers can see the blast radius. Right now the diff only shows this PR's yaml, so we can't confirm nothing else (e.g. status checks, dashboards) depended on the old wording.

kcl/lib/steps/k8s/runkperf.k
```diff
@@ -1,4 +1,5 @@
 import azure_pipelines.ap.steps
+import lib.steps.azure
```
wonderyl: I'm an Agent, trained with yaolei@'s past comments.

`lib/steps/k8s/` importing `lib/steps/azure/` introduces a back-reference from the k8s layer into the cloud layer. Most other k8s steps look cloud-agnostic on purpose. Is that the layering we want? If `CollectKperfResult` truly needs `az aks show`, it might belong under `lib/steps/azure/` (e.g. `azure/kperf.k`) rather than `k8s/`.

kcl/lib/steps/kperf/runkperf.k
```diff
@@ -78,3 +79,73 @@ echo "Ingest complete: $completed ${dataType}s created."
...
+
+sudo -E $(command -v runkperf) -v 3 bench \\
+  --kubeconfig ~/.kube/config \\
+  --runner-image="${runnerImage}" \\
+  --runner-flowcontrol="${flowcontrol}" \\
+  --result /tmp/tmp_results.json \\
+  ${subcmd} ${subcmdArgs}
+"""
+        displayName = "Run kperf benchmark (${subcmd}, flowcontrol=${flowcontrol})"
+    }
```
wonderyl: I'm an Agent, trained with yaolei@'s past comments.

`RunKperfBenchmark` writes `/tmp/tmp_results.json` and `CollectKperfResult` reads it; similarly `CollectKperfResult` writes `/tmp/run-result.json` for the downstream `UploadResult`. These implicit file-path contracts are invisible to a future reader of `lib/steps/k8s/runkperf.k`. At minimum document the paths in the doc comments; better, take the path as a parameter so callers can compose them safely.

kcl/lib/steps/k8s/runkperf.k
```diff
@@ -78,3 +79,73 @@ echo "Ingest complete: $completed ${dataType}s created."
...
+    cloud: $cloud,
+    region: $region,
+    fqdn: $fqdn,
+    private_fqdn: $private_fqdn,
+    k8s_version: $k8s_version,
+    sku_tier: $sku_tier,
+    subscription: $subscription,
+    cluster_id: $cluster_id
+  }')
+
```
wonderyl: I'm an Agent, trained with yaolei@'s past comments.

Two issues with this `errors_raw_data.json` branch:

1. The file is written to `/tmp/errors_raw_data.json` but only `/tmp/run-result.json` is uploaded by the subsequent `UploadResult` step in the pipeline. So the error-rate branch is effectively dead code today — please either upload it too, or remove the block.
2. `error_rate_too_high` triggers whenever `total < count * loadProfile.spec.total` (i.e. *any* loss > 0), which doesn't match the name ("too high"). Pick a real threshold or rename the variable.

kcl/lib/steps/k8s/runkperf.k
```diff
@@ -78,3 +79,73 @@ echo "Ingest complete: $completed ${dataType}s created."
...
+    | jq -cr ".result.errors | map(. + {run_id: \\"$(RUN_ID)\\", cloud: \\"azure\\"}) | .[]" > /tmp/errors_raw_data.json
+  echo "Error raw data:"
+  cat /tmp/errors_raw_data.json
+fi
+
+# Collect final result — pack everything into the "result" field so Kusto ingests it all
+cat /tmp/tmp_results.json \\
+  | jq 'del(.result.errors)' \\
+  | jq -c --argjson ci "$cloud_info" '{ cloud: $ci.cloud, region: $ci.region, fqdn: $ci.fqdn, private_fqdn: $ci.private_fqdn, k8s_version: $ci.k8s_version, sku_tier: $ci.sku_tier, subscription: $ci.subscription, cluster_id: $ci.cluster_id, flowcontrol: "${flowcontrol}", subcmdArgs: "${subcmdArgs}" } + .result | { result: . }' \\
+  | jq -c '. + { timestamp: "'"$TIMESTAMP"'", run_id: "$(RUN_ID)", run_url: "$(System.TeamFoundationCollectionUri)$(System.TeamProject)/_build/results?buildId=$(Build.BuildId)", pipeline: "$(Build.DefinitionName)" }' \\
```
wonderyl: I'm an Agent, trained with yaolei@'s past comments.

All this `jq` reshaping packs fields into `result` so Kusto ingests them — same pattern I flagged on !15480739 (`pipeline.k:137`): "why not calculate the summary with Kusto instead of reinventing a new way to aggregate." This kind of munging belongs in a Kusto function / update policy where we own it once, not duplicated in every benchmark pipeline. At minimum, leave a TODO so we don't enshrine this contract across more pipelines.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -0,0 +1,152 @@
...
+        stage = flowcontrolLabel
+        displayName = flowcontrolLabel
+        jobs = [makeBenchmarkJob(flowcontrol, flowcontrolLabel)]
+    }
+}
+
+output = ap.Pipeline {
+    name = "API Server List Configmaps"
+
+    trigger = "none"
```
wonderyl: I think default of trigger is none, you don't need to set it.

kcl/apiserver_benchmark/configmaps100/pipeline.k
wonderyl: bad agent.
Constant used more than once is preferred. Only single use constants are discouraged.

kcl/lib/steps/k8s/runkperf.k
LeonardCareer: will test it.

kcl/apiserver_benchmark/configmaps100/pipeline.k
wonderyl: dedup with https://github.com/Azure/telescope/pull/1184/changes#r3270890648

kcl/apiserver_benchmark/configmaps100/pipeline.k
wonderyl: Bad agent.
The comment is fair, but not a hard requirement. sometime a bit redundancy improve readability. When it's extracted as a variable to the top, it's harder to find out what the constant is.

kcl/apiserver_benchmark/configmaps100/pipeline.k
wonderyl: Bad agent.
Return type of a lambda is required.

kcl/apiserver_benchmark/configmaps100/pipeline.k
wonderyl: This could be a question to @xinWeiWei24  and @fuweid, do we still need to run this everyday?

kcl/apiserver_benchmark/configmaps100/pipeline.k
wonderyl: Bad agent.
It's only 2 items, good list it out.

kcl/lib/steps/azure/resource_group.k
wonderyl: Suggestion 2 is a good one.

kcl/lib/steps/k8s/runkperf.k
wonderyl: This is a great comment.

kcl/lib/steps/k8s/runkperf.k
wonderyl: Bad agent.
jq was used to format the result, which is useful to get the json clean.

kcl/apiserver_benchmark/configmaps100/pipeline.k
liyu-ma: yes lets also update example pipeline.

kcl/apiserver_benchmark/configmaps100/pipeline.k
LeonardCareer: Thanks, will fix this in next commit

kcl/apiserver_benchmark/configmaps100/pipeline.k
LeonardCareer: make sense, will change it back to az aks create, and util.escapeStr and requestBody can be omitted.

kcl/apiserver_benchmark/configmaps100/pipeline.k
LeonardCareer: yeah, will include this suggestion in next commit, thank you

kcl/lib/steps/azure/resource_group.k
LeonardCareer: yeah, remove the run_id and keep the subscription. It will be included in next commit

kcl/lib/steps/k8s/runkperf.k
```diff
@@ -78,3 +79,73 @@ echo "Ingest complete: $completed ${dataType}s created."
         condition = "and(succeeded(), ne('${totalGiB}', '0'))"
     }
 }
+
+# RunKperfBenchmark runs runkperf bench with the given subcommand and flow control settings.
+# Results are written to /tmp/tmp_results.json for collection by CollectKperfResult.
+RunKperfBenchmark = lambda runnerImage: str, flowcontrol: str, subcmd: str, subcmdArgs: str -> steps.Step {
```
wonderyl: nit: it's useful to put important parameters forward, meaning the one the users must specify. in this specific case, image, subcmd, subcmdArgs, while I feel flowcontrol can be ignored, maybe with a default value.

kcl/apiserver_benchmark/configmaps100/pipeline.k
xinWeiWei24: I think it can be changed to once a week.

kcl/lib/steps/azure/resource_group.k
LeonardCareer: Wow, very good point, didn't think about regenerating all other yaml files. Thanks for the reminder! will add in next commit

Also, your bot is really well trained 👍

kcl/apiserver_benchmark/configmaps100/pipeline.k
LeonardCareer: Do you have a preferred start time?

kcl/lib/steps/k8s/runkperf.k
LeonardCareer: It works well without sudo. remove it.

kcl/apiserver_benchmark/configmaps100/pipeline.k
LeonardCareer: I put 3am UTC every Sunday now. We can get the result at every Monday morning. Happy to change it if any preferred time.

kcl/apiserver_benchmark/configmaps100/pipeline.k
wonderyl, line 54: Why removing WaitForClusterSucceeded?

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -93,30 +66,27 @@ makeBenchmarkJob = lambda flowcontrol: str, flowcontrolLabel: str -> job.Job {
                 "runner",
                 "Standard_D16s_v3",
                 3),
```
wonderyl: Why removing WaitForNodePoolSucceeded?

kcl/lib/steps/kperf/runkperf.k
```diff
@@ -81,71 +80,23 @@ echo "Ingest complete: $completed ${dataType}s created."
...
-  --result /tmp/tmp_results.json \\
-  ${subcmd} ${subcmdArgs}
+  --result ${resultPath} \\
+  ${benchmarkSubcmd} ${benchmarkSubcmdArgs}
 """
-        displayName = "Run kperf benchmark (${subcmd}, flowcontrol=${flowcontrol})"
+        displayName = "Run kperf benchmark (${benchmarkSubcmd}, flowcontrol=${flowcontrol})"
+        target = target
     }
 }
```
wonderyl: Even though this doesn't affect how it's used. I do feel related functions should be organized closely. E.g. in the same file or folder. In this case runkperf and collectKperfResult is pretty much guaranteed to be use together. Then putting them together would help users to discover them easily.

LeonardCareer: Agree. The reason they're currently separated is because CollectKperfResult calls `az aks show`, which requires AzCli. Let me think whether we can split it out.

LeonardCareer: How about this: move the `az aks show` out as an inline azure.AzCli step in the pipeline and generate a file, then refactor CollectKperfResult to read the cloud info from a file. This removes the AzCli dependency so it can live next to RunKperfBenchmark in k8s/. Let me try this one in next commit.

LeonardCareer: Already split `az aks show` out, add the [CollectCloudInfoScript](https://github.com/Azure/telescope/pull/1184/changes#diff-72465da398c24568bcbe6f083b378fa2ee332bd4444b5843770a44989019e6eaR33) and test good, please let me know if any suggestion.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -52,6 +76,11 @@ makeBenchmarkJob = lambda flowcontrol: str, flowcontrolLabel: str -> job.Job {
                 const.DEFAULT_SERVICE_CONNECTION,
                 "Create cluster",
                 createClusterScript),
+            azure.WaitForClusterSucceeded(
```
wonderyl: I thought you said this is not needed, because aks cluster create will block until it's succeeded

LeonardCareer: Yeah, just a verification. removed.

kcl/apiserver_benchmark/configmaps100/pipeline.k
```diff
@@ -74,13 +109,14 @@ makeBenchmarkJob = lambda flowcontrol: str, flowcontrolLabel: str -> job.Job {
                 flowcontrol = flowcontrol,
                 target = "kperf",
                 resultPath = "$(Pipeline.Workspace)/tmp_results.json"),
-            azure.CollectKperfResult(
+            azure.AzCli(
```
wonderyl: why splitting into 2 steps? do you foresee a scenario where you only use one of these steps?

LeonardCareer: We went through three iterations:
RunKperfBenchmark need k8s. CollectKperfResult need to use AzCli (aks show) to collect info.

- First version — CollectKperfResult in k8s/runkperf.k, importing AzCli from azure/. Functions live together, but k8s/ depends on azure/.

- Second version — Moved CollectKperfResult to azure/ since it depends on AzCli. Clean module boundaries, but it is far from RunKperfBenchmark. Bad for user to understand.

- Current version — Split out az aks show as an inline AzCli step in the pipeline, CollectKperfResult becomes pure jq and moves back to k8s/. They live together with no cross-module dependency.  That's why it split into 2 steps.

If you want to change back to previous version or any suggestion, very welcome.

wonderyl: Even though collect cloud info is cloud specific. I feel it's more closely related to Kperf than the cloud, since it's always used with collecting kperf result. You can have a kperf folder instead, and put all these into that folder. We don't need to put everything related to azure to the azure folder. I guess the agent made a wrong recommendation after all.

wonderyl (approved):

PR #1185, opened 2026-05-19
feat(crud): add AKSMachineClient batch scale path (BatchPutMachine)

karenychen:
## Why

Adds the AKS Machine API **batch** scale path (`use_batch_api=True`) to `AKSMachineClient`, dispatching across multiple workers via the `BatchPutMachine` HEADER pattern. This is the final piece needed to scale machine-mode agent pools efficiently at 1000+ machine counts where N individual PUTs would exhaust per-machine RP overhead.

Stack: builds on #1179 (scaffolding) + #1181 (non-batch scale path).

## What

`scale_machine(use_batch_api=True)` now dispatches to a new `_scale_machine_batch` helper instead of raising `NotImplementedError`. The helper:

1. Validates `scale_machine_count % machine_workers == 0` (raises `ValueError` otherwise — see *contract* below).
2. Computes `per_worker = scale_machine_count // machine_workers` and slices the pre-built `names` list by `worker_id` using arithmetic indexing.
3. Submits one PUT per worker in a `ThreadPoolExecutor` (`max_workers=machine_workers`); each PUT carries a contiguous slice of machines in its `BatchPutMachine` header.
4. Isolates per-worker failures: an exception in one worker is logged and that worker's slice is excluded from `successful_machines`; the other workers' results still surface.

### Contract: exact-multiple partitioning (ado-telescope parity)

`scale_machine_count` MUST be an exact multiple of `machine_workers` when `use_batch_api=True`. This mirrors ado-telescope's batch contract verbatim: every worker handles exactly the same per-worker count, so chunk boundaries are byte-identical across runs and the resulting `chunk_0/chunk_1/...` Kusto correlation keys line up exactly with ado dashboards. Non-exact-multiple inputs raise `ValueError` at the `MachineCRUD` boundary instead of producing a short final chunk that would skew per-chunk latency dashboards.

### Background: the `machineName` (NOT `name`) landmine

The `BatchPutMachine` header payload's `batchMachines` array uses the key **`machineName`** — the wrong key (e.g. `name`) causes the API to silently ignore the extras and create only the URL-pointed machine. This was the root cause of an earlier staging build where `succeeded=true` was reported but only `worker_count` machines actually landed.

### 429 retry budget

`_make_batch_request` retries on HTTP 429 with bounded exponential backoff: up to `_BATCH_429_MAX_RETRIES` total attempts (5 by default) with sleeps of 1s, 2s, 4s, 8s between them. All other non-2xx responses raise immediately.

## Metadata enrichment

`scale_machine` records a single `op.add_metadata("command_execution_time", <wall_seconds>)` covering both the individual and batch dispatch paths (uniform shape across modes for downstream dashboards). `successful_machines` / `percentile_node_readiness_times` / `node_readiness_time` / `cluster_info` semantics are unchanged from #1181.

## Verification

- `pylint modules/python/clients/aks_machine_client.py modules/python/tests/test_aks_machine_client.py`: **10.00/10**
- `pytest modules/python/tests/test_aks_machine_client.py`: **27 passed**
- Full module `pytest --cov=. --cov-fail-under=80`: **720 passed, 83.43%** total

## Stack

This is the final code-level slice. Remaining slices in the series:

| # | PR title | Files | LOC |
|---|---|---|---|
| 4 | `feat(crud): add MachineCRUD orchestrator` | `crud/azure/machine_crud.py` + tests | ~200 |
| 5 | `feat(crud): add create-machine + scale-machine CLI subcommands` | `crud/main.py` edits + tests | ~280 |
| 6 | `feat(crud-machine): scenario terraform inputs` | tfvars + test inputs | ~51 |
| 7 | `feat(crud-machine): pipeline templates + production pipeline` | YAML | ~170 |

### DIFF
--- a/modules/python/clients/aks_machine_client.py
+++ b/modules/python/clients/aks_machine_client.py
@@ -769,6 +769,6 @@ class AKSMachineClient(AKSClient):
             backoff *= 2
         text = last_resp.text[:500] if last_resp is not None else ""
         raise RuntimeError(
-            f"batch request exceeded {_BATCH_429_MAX_RETRIES} 429 retries "
-            f"[{ctx}]; last text={text}"
+            f"batch request exceeded {_BATCH_429_MAX_RETRIES} attempts "
+            f"after HTTP 429 responses [{ctx}]; last text={text}"
         )
--- a/modules/python/clients/aks_machine_client.py
+++ b/modules/python/clients/aks_machine_client.py
@@ -704,7 +704,13 @@ class AKSMachineClient(AKSClient):
         # the individual machine PUT site.
         put_timeout = min(request.timeout, _PER_REQUEST_TIMEOUT_CAP)
         self._make_batch_request(
-            "PUT", url, body, put_timeout, batch_header_value=batch_header_value,
+            "PUT",
+            url,
+            body,
+            put_timeout,
+            batch_header_value=batch_header_value,
+            chunk_idx=chunk_idx,
+            first_machine_name=first_machine_name,
         )
         return list(chunk)
 
@@ -715,6 +721,8 @@ class AKSMachineClient(AKSClient):
         data: Dict[str, Any],
         timeout: int,
         batch_header_value: str,
+        chunk_idx: Optional[int] = None,
+        first_machine_name: Optional[str] = None,
     ) -> None:
         """Send a ``BatchPutMachine``-headered request with bounded 429 backoff.
 
@@ -725,6 +733,12 @@ class AKSMachineClient(AKSClient):
         ``_BATCH_429_INITIAL_BACKOFF_SECONDS`` and doubling each attempt
         (1s, 2s, 4s by default). All other non-2xx responses raise immediately.
         """
+        # Compact prefix so every error/log line is self-identifying in Kusto
+        # (chunk_idx + first machine name pinpoint the failing slice without
+        # cross-referencing the per-worker log).
+        ctx = (
+            f"chunk={chunk_idx} target={first_machine_name} {method} {url}"
+        )
         backoff = _BATCH_429_INITIAL_BACKOFF_SECONDS
         last_resp: Optional[requests.Response] = None
         for attempt in range(_BATCH_429_MAX_RETRIES):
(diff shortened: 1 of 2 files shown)


modules/python/clients/aks_machine_client.py
```diff
@@ -573,3 +580,174 @@ def _create_single_machine(self, name: str, request: SimpleNamespace) -> bool:
...
+    def _create_batch_machines(
+        self,
+        request: SimpleNamespace,
+        chunk: List[str],
+        chunk_idx: int,
+    ) -> List[str]:
+        """Submit a batch-PUT request creating every machine in ``chunk``.
+
+        Uses the AKS Machine API ``BatchPutMachine`` HEADER pattern (NOT the
+        non-existent ARM ``/$batch`` envelope): a single PUT to the first
```
karenychen:
```suggestion
        Uses the AKS Machine API ``BatchPutMachine`` HEADER pattern (implicit
        contract with AKS NAP, not publicly available): a single PUT to the first
```

modules/python/clients/aks_machine_client.py
```diff
@@ -446,7 +449,11 @@ def scale_machine(
                 names = [
                     f"{prefix}-machine-{i + 1}" for i in range(scale_machine_count)
                 ]
-                successful = self._scale_machine_individually(request, names)
+                if use_batch_api:
+                    successful, batch_times = self._scale_machine_batch(request, names)
+                    op.add_metadata("batch_command_execution_times", batch_times)
+                else:
```
karenychen, line 457: we should add `command_execution_time` to both cases, unify them into 1 `add_metadata()` call

modules/python/clients/aks_machine_client.py
```diff
@@ -573,3 +582,176 @@ def _create_single_machine(self, name: str, request: SimpleNamespace) -> bool:
...
+        """
+        backoff = _BATCH_429_INITIAL_BACKOFF_SECONDS
+        last_resp: Optional[requests.Response] = None
+        for attempt in range(_BATCH_429_MAX_RETRIES):
+            headers = {
+                "Authorization": f"Bearer {self._get_access_token()}",
+                "Content-Type": "application/json",
+                "BatchPutMachine": batch_header_value,
+            }
+            resp = requests.request(
```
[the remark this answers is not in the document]
karenychen: Switched `_make_batch_request` to `self._session.request(...)` so the batch path reuses the client's pooled connections (and any adapter/proxy/timeout configured on the Session) -- same as `make_request`. Done in 3223e5a9.

vittoriasalim (approved):

PR #1192, opened 2026-05-25
Add a script to split a large ado pipeline yaml into smaller pieces

wonderyl:
Ado pipeline allows up to 2MB of yaml file. To overcome the limit, I created this script that using greedy algo to split out a subset of the yaml into templates.
The algo is by no means optimal, we can optimize it later.


scripts/testdata/split_top_level_jobs/expected/pipeline_0.yaml
```diff
@@ -0,0 +1,16 @@
+name: Test Pipeline
+pool: TestPool
+trigger:
+- main
+jobs:
+- template: $(Pipeline.Workspace)/s/pipeline_1.yaml
```
vittoriasalim: will this call jobs header twice

Image

```
jobs: 
jobs: 
 - job: job1
```
instead of 
```
jobs:
 - job: job1
 ```

wonderyl: This is actually how template works. I tried it with the test pipeline. Here's the run https://dev.azure.com/akstelescope/telescope/_build/results?buildId=68417&view=results
Here's the code. https://github.com/Azure/telescope/blob/yaolei/test/pipelines/system/new-pipeline-test.yml

However, thanks to your comment, I noticed the template path is relative to the yaml file. I need to fix that.

vittoriasalim (approved):

PR #1196, opened 2026-05-27
feat: extend NodePool schema and CreateResourceGroup with optional fields

xinWeiWei24:
Add vnetSubnetId and osSKU optional fields to NodePool schema, and tags optional parameter to CreateResourceGroup. Both default to empty/omitted for backwards compatibility with existing pipelines.

### DIFF
--- a/kcl/lib/steps/azure/create_node_pool.k
+++ b/kcl/lib/steps/azure/create_node_pool.k
@@ -9,12 +9,16 @@ schema NodePool:
     taintPrefix: str = ""
     labels: str = ""
     maxPods?: int
+    vnetSubnetId: str = ""
+    osSKU: str = ""
 
 CreateNodePool = lambda serviceConnection: str, cluster: str, resourceGroup: str, subscription: str, pool: NodePool, noWait: bool = False -> steps.Step {
     flagDecision = [
         "--node-taints ${pool.taintPrefix}=true:NoSchedule" if pool.taintPrefix else "",
         "--labels \"${pool.labels}\"" if pool.labels else "",
         "--max-pods ${pool.maxPods}" if pool.maxPods else "",
+        "--vnet-subnet-id \"${pool.vnetSubnetId}\"" if pool.vnetSubnetId else "",
+        "--os-sku ${pool.osSKU}" if pool.osSKU else "",
         "--no-wait" if noWait else "",
     ]
     flags = " \\\n  ".join([f for f in flagDecision if f])
--- a/kcl/lib/steps/azure/create_node_pool.k
+++ b/kcl/lib/steps/azure/create_node_pool.k
@@ -9,8 +9,8 @@ schema NodePool:
     taintPrefix: str = ""
     labels: str = ""
     maxPods?: int
-    vnetSubnetId: str = ""
-    osSKU: str = ""
+    vnetSubnetId?: str
+    osSKU?: str
 
 CreateNodePool = lambda serviceConnection: str, cluster: str, resourceGroup: str, subscription: str, pool: NodePool, noWait: bool = False -> steps.Step {
     flagDecision = [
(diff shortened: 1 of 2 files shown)


kcl/lib/steps/azure/create_node_pool.k
```diff
@@ -8,11 +8,15 @@ schema NodePool:
     count: int
     taintPrefix: str = ""
     labels: str = ""
+    vnetSubnetId: str = ""
```
liyu-ma: Is it better use `str?` type which has a default value `None`

xinWeiWei24: addressed

kcl/lib/steps/azure/create_node_pool.k
```diff
@@ -8,11 +8,15 @@ schema NodePool:
     count: int
     taintPrefix: str = ""
     labels: str = ""
+    vnetSubnetId?: str
+    osSKU?: str
 
 CreateNodePool = lambda serviceConnection: str, cluster: str, resourceGroup: str, subscription: str, pool: NodePool, noWait: bool = False -> steps.Step {
```
wonderyl, line 15: The list of parameters becomes longer and longer. I wonder if it makes sense to let users to write a `az aks` command directly instead? The wrapper is only meaningful if it simplify things. now it's just pass through everything.

kcl/lib/steps/azure/resource_group.k
```diff
@@ -1,8 +1,12 @@
 import azure_pipelines.ap.steps
 
-CreateResourceGroup = lambda serviceConnection: str, name: str, location: str, subscription: str -> steps.Step {
+CreateResourceGroup = lambda serviceConnection: str, name: str, location: str, subscription: str, tags: str = "" -> steps.Step {
+    tagsFlag = " \\\n  --tags ${tags}" if tags else ""
```
wonderyl: Normally it's called an argument.

```
$ az aks create -h
The behavior of this command has been altered by the following extension: aks-preview

Command
    az aks create : Create a new managed Kubernetes cluster.

Arguments
    --name -n                                               [Required] : Name of the managed
```

kcl/lib/steps/azure/resource_group.k
```diff
@@ -1,8 +1,12 @@
...
     script = """
-az group create --name "${name}" --location "${location}" --subscription "${subscription}"
+az group create \\
+  --name "${name}" \\
+  --location "${location}" \\
+  --subscription "${subscription}"${tagsFlag}
```
wonderyl: nit: I feel it's more easier to read when you do 
```
az group create \\
  --name "${name}" \\
  --location "${location}" \\
  --subscription "${subscription}" \\n
  ${"--tags " + tags if tags else ""}
```
Even though it will end up have a training \n

kcl/lib/steps/azure/create_node_pool.k
liyu-ma: I feel it is ok as long as we don't add more params to `CreateNodePool` signature. We need `schema NodePool` to represent CL2 pool / worker pool and pass around. Users will need to configure/enrich `NodePool` for their use cases. I learnt you can do something like below which does not feel complicated:
```
  planned = NodePool {
      sku = "Standard_D4"
      count = 3
  }

  configured = NodePool {**planned, vnetSubnetId = "my_subnet_id"}
```

kcl/lib/steps/azure/create_node_pool.k
```diff
@@ -8,11 +8,15 @@ schema NodePool:
     count: int
     taintPrefix: str = ""
     labels: str = ""
+    vnetSubnetId?: str
+    osSKU?: str
```
liyu-ma, line 13: btw what is difference between `sku` and `osSKU`? I added `sku` param to represent the vm size for the node

xinWeiWei24: The OS SKU of the agent node pool. Ubuntu or CBLMariner for Linux. Windows2019 or Windows2022 for Windows.
Allowed values: AzureLinux, CBLMariner, Mariner, Ubuntu, Windows2019, Windows2022.

liyu-ma (approved):

PR #1197, opened 2026-05-27
feat: add identity and role assignment lib steps

xinWeiWei24:
Add CreateManagedIdentity (exports IDENTITY_CLIENT_ID and IDENTITY_ID as ADO variables) and CreateRoleAssignment to lib/steps/azure.


vittoriasalim (approved):

kcl/lib/steps/azure/identity.k
```diff
@@ -0,0 +1,25 @@
...
+  --resource-group "${resourceGroup}" \\
+  --name "${name}" \\
+  --subscription "${subscription}"
+
+IDENTITY_CLIENT_ID=$(az identity show \\
+  --resource-group "${resourceGroup}" \\
+  --name "${name}" \\
+  --subscription "${subscription}" \\
+  --query clientId -o tsv)
+echo "##vso[task.setvariable variable=IDENTITY_CLIENT_ID]$IDENTITY_CLIENT_ID"
```
liyu-ma: The functions here push the value to a hard-coded var `IDENTITY_CLIENT_ID`. Can we take a param to allow user to pass in a env name they want to store the value? This would help if there are multiple identities they want to create.

kcl/lib/steps/azure/identity.k
```diff
@@ -0,0 +1,25 @@
...
+
+IDENTITY_ID=$(az identity show \\
```
liyu-ma: Does `az identity show` return both client id and identity id? I am a bit surprised if you want to call it twice.

kcl/lib/steps/azure/role_assignment.k
```diff
@@ -0,0 +1,12 @@
...
+
+CreateRoleAssignment = lambda serviceConnection: str, scope: str, role: str, assignee: str, subscription: str -> steps.Step {
+    script = """
+az role assignment create \\
+  --scope "${scope}" \\
+  --role "${role}" \\
+  --assignee "${assignee}" \\
+  --subscription "${subscription}"
+"""
+    AzCli(serviceConnection, "Assign role ${role}", script)
```
liyu-ma: Would be good to show assignee as well

kcl/lib/steps/azure/identity.k
xinWeiWei24: fixed

kcl/lib/steps/azure/identity.k
xinWeiWei24: Thanks! now only call once

kcl/lib/steps/azure/role_assignment.k
xinWeiWei24: added

kcl/lib/steps/azure/role_assignment.k
```diff
@@ -0,0 +1,12 @@
+import azure_pipelines.ap.steps
+
+CreateRoleAssignment = lambda serviceConnection: str, scope: str, role: str, assignee: str, subscription: str -> steps.Step {
```
wonderyl: This function doesn't provide much value, would it be easier if the user just call AzCli with the script directly?

wonderyl: When create a function, ask yourself, what does this function encapsulate?

xinWeiWei24: This is intended to be invoked multiple times, since the 3 input variables ( scope, role, assignee) can differ per call. Wrapping it as a function avoids duplicating the same CLI script logic.

kcl/lib/steps/azure/identity.k
```diff
@@ -0,0 +1,18 @@
+import azure_pipelines.ap.steps
+
+CreateManagedIdentity = lambda serviceConnection: str, subscription: str, resourceGroup: str, name: str, exportVar: str = "IDENTITY" -> steps.Step {
```
wonderyl, line 3: IIUC, you always need to create managed identity, then assign role, then bind it to a service account. Does it make sense to encapsulate the 3 steps into one?

xinWeiWei24: Not for my case. I’m not binding it to a service account, and the role assignments target different resources and roles.

liyu-ma (approved):

kcl/lib/steps/azure/role_assignment.k
xinWeiWei24: will delete this and add the step to subnet creation

kcl/lib/steps/azure/identity.k
wonderyl: How do you use the identity?

xinWeiWei24: assign the "netwrok contributor" role to subnet so that it can be used during cluster creation, like this
```bash
az role assignment create --scope "$APISERVER_SUBNET_ID" --role "Network Contributor" --assignee "$IDENTITY_CLIENT_ID" --subscription "$SUBSCRIPTION"
```

PR #1201, opened 2026-05-28
Add optional maxPods to NodePool for az aks nodepool add

liyu-ma:
Allow configuring --max-pods at pool creation


kcl/lib/steps/azure/create_node_pool.k
wonderyl, line 1: will this change any existing pipeline.yaml?

liyu-ma: Updated the exiting pipeline yaml

wonderyl (approved):

PR #1202, opened 2026-05-28
Dipowell/crud jobs v2

diamondpowell:
Add Job workload support to CRUD benchmarking framework

## Summary

Adds Job workload support to the CRUD benchmarking framework - the third and final of three planned workload methods (`deployment`, `statefulset`, `job`). Measures K8s Job create/verify latency on AKS node pools.

## Changes

`modules/python/crud/workload_templates/job.yml`
- New K8s Job manifest template using nginx config validation for quick completion
- Uses `JOB_COMPLETIONS` placeholder, configurable node affinity via label_selector

`modules/python/crud/azure/node_pool_crud.py`
- Add `job` entry to `WORKLOAD_CONFIG` with template path, completions key, `complete` condition, and `verify_pods_ready=False` (jobs terminate after completion)
- Add `create_job()` using unified `_create_workloads` helper

`modules/python/crud/main.py`
- Add `job` subparser with `--completions` argument (distinct from `--replicas`)
- Add `elif command == "job"` routing in `handle_workload_operations`

`modules/python/clients/kubernetes_client.py`
- Add `job`/`jobs` to `valid_conditions` with `['complete', 'failed']`
- Add `_check_job_condition` and `_is_job_condition_met` for job status polling
- Extend `wait_for_condition` to support Job resource type

`steps/engine/crud/k8s/execute.yml`
- Add `job` script block (Azure-only, consistent with deployment/statefulset)

`pipelines/perf-eval/GPU Benchmark/k8s-gpu-cluster-crud.yml`
- Add `completions: 1` to matrix entries

## Tests

`test_azure_node_pool_crud.py`:
- `test_create_job_success` - happy path
- `test_create_job_failure` - all fail to complete
- `test_create_job_no_client` - returns early when k8s client unavailable
- `test_create_job_partial_success` - continues on failures, returns False

`test_kubernetes_client.py`:
- `test_wait_for_condition_job_complete_success`
- `test_wait_for_condition_job_failed`
- `test_wait_for_condition_job_timeout`
- `test_wait_for_condition_job_not_found`

### DIFF
--- a/modules/python/clients/kubernetes_client.py
+++ b/modules/python/clients/kubernetes_client.py
@@ -977,6 +977,8 @@ class KubernetesClient:
             'deployments': ['available', 'progressing', 'replicafailure', 'ready'],
             'statefulset': ['ready'],
             'statefulsets': ['ready'],
+            'job': ['complete', 'failed'],
+            'jobs': ['complete', 'failed'],
         }
 
         # Validate wait_condition_type format and type
@@ -1053,6 +1055,9 @@ class KubernetesClient:
             if resource_type_lower in ['statefulset', 'statefulsets']:
                 return self._check_statefulset_condition(resource_name, namespace, wait_all)
 
+            if resource_type_lower in ['job', 'jobs']:
+                return self._check_job_condition(resource_name, condition_type, namespace, wait_all)
+
             logger.warning(f"Unsupported resource type for condition checking: {resource_type}")
             return False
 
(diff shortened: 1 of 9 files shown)


modules/python/crud/workload_templates/job.yml
```diff
@@ -0,0 +1,20 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: myapp-{{NODE_POOL_NAME}}-{{INDEX}}
```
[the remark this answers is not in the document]
diamondpowell: addresses in commit

modules/python/clients/kubernetes_client.py
```diff
@@ -1114,6 +1119,50 @@ def _is_statefulset_ready(self, statefulset) -> bool:
...
+                if condition.type.lower() == condition_type_lower and condition.status == "True":
+                    return True
+
+        # Fallback: check completion/failure status directly
+        if condition_type_lower == 'complete':
+            spec_completions = job.spec.completions or 1
+            return (job.status.succeeded is not None and
+                    job.status.succeeded >= spec_completions)
+        if condition_type_lower == 'failed':
+            return job.status.failed is not None and job.status.failed > 0
```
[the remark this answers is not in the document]
diamondpowell: addressed in commit

modules/python/clients/kubernetes_client.py
```diff
@@ -1114,6 +1119,48 @@ def _is_statefulset_ready(self, statefulset) -> bool:
...
+                    return True
+
+        # Fallback: infer completion from success counters when conditions are absent
+        if condition_type_lower == 'complete':
+            spec_completions = job.spec.completions or 1
+            return (job.status.succeeded is not None and
+                    job.status.succeeded >= spec_completions)
+
+        return False
+
```
xinWeiWei24, line 1160: do you think we can use wait_for_job_completed instead of adding these two functions. https://github.com/Azure/telescope/blob/48c6e66cfa8b8b88d3aa2f80658f757962b6000f/modules/python/clients/kubernetes_client.py#L372

modules/python/crud/main.py
```diff
@@ -531,6 +546,41 @@ def main():
...
+        default=None,
+        help="Directory containing Kubernetes manifest files"
+    )
+    job_parser.add_argument(
+        "--label-selector",
+        default="app=nginx-container",
+        help="Label selector for created pods (default: app=nginx-container)"
+    )
+    job_parser.set_defaults(func=handle_workload_operations)
+
```
xinWeiWei24, line 562: This job parser can still use workload_common_parser, right? All the arguments are already defined there.

diamondpowell: Yes they are, it didnt include completions but I just added as argument. That is a way better way to do it, so thanks for that catch.

diamondpowell: Updated to use workload_common_parser like deployment and statefulset. The job parser now inherits --replicas which jobs don't use, but it's harmless... the argument is simply ignored in the job handler.

modules/python/clients/kubernetes_client.py
diamondpowell: Well no, only because they are different. So wait_for_job_completed has its own polling loop and is a complete waiter, while job_condition are check-one time helpers (job object/job name). But I refactored it to call _is_job_condition_met so both code paths share the same condition logic now.

modules/python/clients/kubernetes_client.py
```diff
@@ -1114,6 +1119,48 @@ def _is_statefulset_ready(self, statefulset) -> bool:
...
+            for job in jobs:
+                if not self._is_job_condition_met(job, condition_type):
+                    return False
+
+            return True
+
+        except client.rest.ApiException as e:
+            if e.status == 404:
+                logger.debug("Job not found, waiting...")
```
liyu-ma: Since `False` is returned, I assume we are not waiting here right? Then can you remove this debug logging.

diamondpowell: Yea its not waiting and that does look misleading. I had this same pattern for Deployment and StatefulSet also. So I removed all three since the function doesn't actually wait, it just returns False.

liyu-ma (approved):

PR #1208, opened 2026-06-02
Dipowell/node readiness timing

diamondpowell:
## Summary

Adds `node_readiness_time` as a separate metric in the open-source CRUD module to match internal repo behavior. The internal repo captures how long K8s nodes take to become Ready independently from when the ARM API completes. The open-source repo was missing this - it only had combined duration.

Azure API says "done" when the control plane finishes, but nodes might not be schedulable yet. Capturing both timestamps separately enables regression analysis:

- `command_execution_time > node_readiness_time` -> ARM layer is the bottleneck
- `node_readiness_time > command_execution_time` -> K8s layer is the bottleneck

## Changes

**utils/provisioning_instrumentation.py** (new)
- Extracted `instrument_nodepool_provisioning()` - runs ARM and K8s readiness concurrently via `ThreadPoolExecutor`, captures separate timing metadata
- Extracted `begin_create_or_update_with_retry()` - ARM operation with retry on `OperationNotAllowed`/`EtagMismatch`, returns `retry_occurred` boolean

**clients/aks_client.py**
- Thin `_instrument_nodepool_provisioning()` wrapper that builds callables and delegates to the extracted module
- All three call sites (create, scale, progressive scale) now use retry consistently
- Removed `use_retry` parameter
- Net reduction: 1020 -> 958 lines (resolves pylint `too-many-lines`)

**clients/kubernetes_client.py**
- Updated `wait_for_nodes_ready()` docstring to reflect actual return type and parameter names

**tests/utils/test_provisioning_instrumentation.py** (new)
- 8 unit tests covering both extracted functions: timing metadata, retry_occurred flag, error propagation, timeout, retryable vs non-retryable errors

**tests/clients/test_aks_client.py**
- Added `test_create_node_pool_retry_occurred_metadata` for end-to-end retry verification
- Updated time mocks to patch `utils.provisioning_instrumentation.time` for timing assertions

## Timing metadata stored via `op.add_metadata()`:
- `node_readiness_time`: seconds from start until K8s nodes were Ready
- `command_execution_time`: seconds from start until ARM operation completed
- `retry_occurred`: boolean, True if ARM operation hit a transient error and retried

## Implementation notes
- Uses `ThreadPoolExecutor(max_workers=2)` to run ARM polling and K8s readiness checks concurrently
- Both tasks run to completion even if one fails, enabling partial diagnostics
- `retry_occurred` metadata enables filtering skewed `command_execution_time` values from regression analysis
- Extraction keeps `aks_client.py` under 1,000 lines without disabling pylint


modules/python/clients/aks_client.py
```diff
@@ -60,6 +62,92 @@ def _get_operation_context(self):
...
+                    raise arm_result
+                raise k8s_result
+
+            # Both succeeded - unpack k8s result
+            ready_nodes, ready_timestamp = k8s_result
+
+            # Calculate times relative to start
+            node_readiness_time = ready_timestamp - start_time
+            command_execution_time = completion_time - start_time
+
```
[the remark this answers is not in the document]
diamondpowell: Fixed command_execution_time to measure ARM completion independently

modules/python/clients/aks_client.py
```diff
@@ -60,6 +61,85 @@ def _get_operation_context(self):
...
+            arm_response, arm_timestamp = arm_result
+            ready_nodes, ready_timestamp = k8s_result
+
+            # Calculate times relative to start
+            node_readiness_time = ready_timestamp - start_time
+            command_execution_time = arm_timestamp - start_time
+
+            return arm_response, ready_nodes, node_readiness_time, command_execution_time
+
+        return asyncio.run(_run())
```
liyu-ma: asyncio requires event loop. With this implementation it embed an implicit requirement that caller must not be in an event loop already. For example if clients want to invoke `create_node_pool` in an async fashion at higher level, the asyncio.run here will fail as an it is called from within an event loop already. Ideally if we want to use `async` it would be good to propagate the `async` chain all the way up but this is much bigger refactoring.

How about just spinning up a small `ThreadPoolExecutor` similar like this:
```
def _run_concurrent_arm_and_readiness(self, ...):
    with ThreadPoolExecutor(max_workers=2) as ex:
        arm_future = ex.submit(...)
        k8s_future = ex.submit(...)
        wait([arm_future, k8s_future])

```

modules/python/clients/aks_client.py
```diff
@@ -60,6 +61,85 @@ def _get_operation_context(self):
 
         return OperationContext
 
+    def _run_concurrent_arm_and_readiness(
```
liyu-ma: I think `_instrument_nodepool_provisioning` is a better name for the method

modules/python/clients/aks_client.py
```diff
@@ -60,6 +61,85 @@ def _get_operation_context(self):
...
+        self,
+        poller,
```
liyu-ma: I see only `begin_create_or_update ` is passed in so no need to have a separate `poller` param

modules/python/clients/aks_client.py
```diff
@@ -60,6 +61,85 @@ def _get_operation_context(self):
...
+        node_count: int,
+        label_selector: str,
+        start_time: float
```
liyu-ma: I think `start_time` can be captured within this function, no need to be passed from outside

modules/python/clients/aks_client.py
diamondpowell: Good call on the event loop issue. Switched to ThreadPoolExecutor with max_workers=2 and concurrent.futures.wait, same concurrency behavior without the asyncio dependency.

modules/python/clients/aks_client.py
diamondpowell: Done, much clearer name for what this method actually does.

modules/python/clients/aks_client.py
diamondpowell: Moved the begin_create_or_update call inside the method. Callers now just pass node_pool_name, cluster_name, parameters, and node_count, which cleaned up all 3 call sites pretty nicely.

modules/python/clients/aks_client.py
diamondpowell: Rolled this into the same change as the poller removal since both touched the method signature. start_time and label_selector are both derived internally now.

modules/python/clients/aks_client.py
```diff
@@ -448,27 +538,26 @@ def create_node_pool(
                         gpu_instance_profile=gpu_instance_profile,
                         gpu_mig_strategy=gpu_mig_strategy,
                     )
+                    command_execution_time = time.time() - start_time
```
liyu-ma: What is the reason for the addition here? Is this your PR's scope to instrument GPU pool?

modules/python/clients/aks_client.py
```diff
@@ -61,6 +62,94 @@ def _get_operation_context(self):
...
+                return_timestamp=True,
+            )
+
+        with ThreadPoolExecutor(max_workers=2) as executor:
+            arm_future = executor.submit(_poll_arm)
+            k8s_future = executor.submit(_wait_k8s)
+
+        arm_exc = arm_future.exception()
+        k8s_exc = k8s_future.exception()
+        k8s_exc = k8s_future.exception()
```
liyu-ma: Why the duplication for line 116 and 117?

modules/python/clients/aks_client.py
```diff
@@ -61,6 +62,94 @@ def _get_operation_context(self):
...
+        arm_response, arm_timestamp = arm_future.result()
+        ready_nodes, ready_timestamp = k8s_future.result()
+
+        # Calculate times relative to start
+        node_readiness_time = ready_timestamp - start_time
+        command_execution_time = arm_timestamp - start_time
+
+        return arm_response, ready_nodes, node_readiness_time, command_execution_time
+
+    def _log_timing_metrics(self, op, node_pool_name, node_readiness_time, command_execution_time):
```
liyu-ma: How about inline this function within `_instrument_nodepool_provisioning` and call it at the end?

modules/python/clients/aks_client.py
```diff
@@ -61,6 +62,94 @@ def _get_operation_context(self):
...
+        label_selector = f"agentpool={node_pool_name}"
+
+        poller = self.aks_client.agent_pools.begin_create_or_update(
+            resource_group_name=self.resource_group,
+            resource_name=cluster_name,
+            agent_pool_name=node_pool_name,
+            parameters=parameters,
+        )
+
+        def _poll_arm():
```
liyu-ma: I don't see the need to have separate `_poll_arm` and `_wait_k8s`. How about simplifying it with lamda, for example:
```
        with ThreadPoolExecutor(max_workers=2) as executor:
            arm_future = executor.submit(lambda: (poller.result(), time.time()))
            k8s_future = executor.submit(
                lambda: self.k8s_client.wait_for_nodes_ready(
                    node_count=node_count,
                    operation_timeout_in_minutes=self.operation_timeout_minutes,
                    label_selector=label_selector,
                    return_timestamp=True,
                )
            )
```

modules/python/clients/aks_client.py
```diff
@@ -608,27 +697,21 @@ def scale_node_pool(
                 node_pool.count = node_count
 
                 logger.info(f"Scaling node pool {node_pool_name} to {node_count} nodes")
-                self._begin_update_with_retry(
```
liyu-ma, line 691: The deletion here is a behavior change. Your `_instrument_nodepool_provisioning` calls `begin_create_or_update` which does not do retry. I would suggest using `_begin_update_with_retry` within your instrumentation as it retries on transient errors - but this may skew your metrics.

modules/python/clients/aks_client.py
```diff
@@ -836,23 +919,15 @@ def _progressive_scale(
                         "cluster_info", self.get_cluster_data(cluster_name)
                     )
                     node_pool.count = step  # Update node count in the node pool object
-                    self._begin_update_with_retry(
```
liyu-ma, line 927: Same as above. Now the new codes lose the retry support.

modules/python/clients/kubernetes_client.py
```diff
@@ -237,15 +237,16 @@ def delete_node(self, node_name):
             else:
                 raise Exception(f"Error deleting Node '{node_name}': {str(e)}") from e
 
-    def wait_for_nodes_ready(self, node_count, operation_timeout_in_minutes, label_selector=None):
+    def wait_for_nodes_ready(self, node_count, operation_timeout_in_minutes, label_selector=None, return_timestamp=False):
```
liyu-ma: The new param makes this function complicated and more importantly - it is unnecessary. You should be able to capture the current time at call site, something like:
```
k8s_future = executor.submit(
                lambda: (
                    self.k8s_client.wait_for_nodes_ready( ... ),
                    time.time(),
                )
            )
```
The `time.time()` function will only be executed when wait_for_nodes_ready() completes.

modules/python/clients/aks_client.py
```diff
@@ -61,6 +62,94 @@ def _get_operation_context(self):
...
+        command_execution_time = arm_timestamp - start_time
+
+        return arm_response, ready_nodes, node_readiness_time, command_execution_time
+
+    def _log_timing_metrics(self, op, node_pool_name, node_readiness_time, command_execution_time):
+        """Log timing metrics with bottleneck analysis and add metadata to operation."""
+        op.add_metadata("node_readiness_time", node_readiness_time)
+        op.add_metadata("command_execution_time", command_execution_time)
+        delta = abs(command_execution_time - node_readiness_time)
+        bottleneck = "ARM" if command_execution_time > node_readiness_time else "K8s"
```
liyu-ma: Don't like the way to log a 'bottleneck' - it is very subjective. Logging should better just record the facts e.g timestamps, delta. Your data analytics tool is responsible for decision making whether it is a bottleneck

modules/python/clients/aks_client.py
diamondpowell: I went head and removed. The original thought was to capture timing for all create paths so the metadata would be consistent, but agreed it's out of scope for this PR.

modules/python/clients/aks_client.py
diamondpowell: Ahh that is a miss on my part, removed the duplicate.

modules/python/clients/aks_client.py
diamondpowell: Ah thats a nice idea. Replaced _poll_arm and _wait_k8s with inline lambdas. Each lambda wraps its operation and pairs the result with a time.time() call, so the timestamp is captured right when the operation completes. The ARM lambda returns (poller.result(), time.time()) and the K8s lambda returns (wait_for_nodes_ready(...), time.time()). Keeps everything in one place instead of jumping to separate helper methods. Making it simpler like this is a good idea.

modules/python/clients/aks_client.py
diamondpowell: Yeah, my initial refactor, I dropped it when I moved the ARM call inside _instrument_nodepool_provisioning.  So went ahead and added a use_retry parameter so when True, the method calls _begin_update_with_retry instead of begin_create_or_update. scale_node_pool and _progressive_scale both pass use_retry=True now, so the transient error handling is restored.

modules/python/clients/aks_client.py
diamondpowell: Yes this is fixed now.

modules/python/clients/kubernetes_client.py
diamondpowell: Oh yes. Agreed, it is cleaner this way. I removed the parameter entirely and capture time.time() in the lambda at the call site instead.

modules/python/clients/aks_client.py
diamondpowell: That's fair. Removed the bottleneck label - now just logs ARM time, K8s time, and the absolute delta. Analysis stays in the query layer.

modules/python/clients/aks_client.py
diamondpowell: Done. I went ahead and moved the logging and metadata recording to the end of _instrument_nodepool_provisioning directly. No separate helper anymore.

diamondpowell:
Pylint `too-many-lines` (C0302) is failing because `aks_client.py` was at 999 lines on main after #1233. This PR adds 60 net lines for the instrumentation method. Both `eks_client.py` and `kubernetes_client.py` already disable this rule for the same reason. Something to think about for future additions to this file.

modules/python/clients/aks_client.py
```diff
@@ -519,23 +609,17 @@ def create_node_pool(
...
-
-                # Wait for nodes to be ready
-                ready_nodes = self.k8s_client.wait_for_nodes_ready(
-                    node_count=node_count,
-                    operation_timeout_in_minutes=self.operation_timeout_minutes,
-                    label_selector=label_selector,
-                )
+                    # Run ARM and K8s readiness concurrently to capture both timings
+                    ready_nodes = self._instrument_nodepool_provisioning(
+                        node_pool_name, cluster_name, parameters, node_count, op
```
liyu-ma, line 521: I see you are skipping retry here (which is consistent with original impl). Do you mind doing a favor here updating it to use retry as well? The `OperationNotAllowed` exception can happen either when creating or updating a pool. Then we can remove the `use_retry` param from `_instrument_nodepool_provisioning` method.

modules/python/clients/aks_client.py
```diff
@@ -61,6 +62,95 @@ def _get_operation_context(self):
...
+            )
+            if arm_exc:
+                raise arm_exc
+            raise k8s_exc
+
+        _, arm_timestamp = arm_future.result()
+        ready_nodes, ready_timestamp = k8s_future.result()
+
+        node_readiness_time = ready_timestamp - start_time
+        command_execution_time = arm_timestamp - start_time
```
liyu-ma: Your `command_execution_time` value would be skewed if retry occurs (wait intervals + all retry latencies). IMO it would be good to capture if retry occurred here as metadata then you can filter it out from your results. If you want to do that one option could be adding `retry_occurred` as a boolean return value from `_begin_update_with_retry`. But just a suggestion as it is your metrics :)

modules/python/clients/aks_client.py
diamondpowell: Good call, done. All paths go through retry now and I dropped the use_retry param entirely. Also extracted the retry + concurrent timing logic into a separate provisioning_instrumentation.py module to get aks_client.py back under 1,000 lines.

modules/python/clients/aks_client.py
diamondpowell: Agreed, added retry_occurred as a bool from begin_create_or_update_with_retry. It's always emitted now so downstream queries have a consistent schema. Appreciate the suggestion.

liyu-ma (approved):

PR #1212, opened 2026-06-05
Fail fast on terminal AKS machine failures

karenychen:
## Summary
- add ListMachines polling during machine readiness waits to fail early when expected machines terminally fail
- keep BatchPutMachine request sizing within the 50-machine service limit
- record successful_machines as a count instead of uploading machine names
- add focused unit coverage for timeout propagation, ListMachines pagination, terminal failure detection, and batch sizing

## Validation
- python3 -m pytest modules/python/tests/test_aks_machine_client.py modules/python/tests/test_machine_crud.py modules/python/tests/test_crud_main.py -q
- python3 -m py_compile modules/python/clients/aks_machine_client.py modules/python/tests/test_aks_machine_client.py
- PYTHONPATH=modules/python python3 -m pylint modules/python/tests/test_aks_machine_client.py modules/python/clients/aks_machine_client.py modules/python/crud/azure/machine_crud.py modules/python/crud/main.py
- git diff --check

### DIFF
--- a/modules/python/clients/aks_machine_client.py
+++ b/modules/python/clients/aks_machine_client.py
@@ -59,7 +59,7 @@ _BATCH_MAX_MACHINES_PER_REQUEST = 50
 _MACHINE_TERMINAL_STATES = {"Succeeded", "Failed", "Canceled"}
 _MACHINE_FAILURE_STATES = {"Failed", "Canceled"}
 _MACHINE_FAILURE_DETAIL_LIMIT = 10
-_LIST_MACHINES_PAGE_LIMIT = 50
+_LIST_MACHINES_MAX_PAGES = 50
 _MACHINE_FAILURE_CHECK_INTERVAL_SECONDS = 30
 _NODE_READINESS_POLL_INTERVAL_SECONDS = 2
 # urllib3's default ``HTTPAdapter`` (mounted implicitly by ``requests.Session``)
@@ -79,12 +79,12 @@ class MachineProvisioningFailed(RuntimeError):
     """Terminal Machine failure with readiness metadata captured before failure."""
     def __init__(
         self,
-        agentpool_name: str,
+        agent_pool_name: str,
         failed_machines: List[Dict[str, Any]],
         readiness_envelope: Dict[str, Dict[str, Any]],
     ):
         super().__init__(
-            f"machine provisioning failed in agentpool {agentpool_name}: "
+            f"machine provisioning failed in agentpool {agent_pool_name}: "
             f"failed_machines={failed_machines}"
         )
         self.failed_machines = failed_machines
@@ -164,7 +164,7 @@ class AKSMachineClient(AKSClient):
     # ---- Machine API: agent pool provisioning ----
     def create_machine_agentpool(
         self,
-        agentpool_name: str,
+        agent_pool_name: str,
         vm_size: str,
         cluster_name: Optional[str] = None,
         timeout: int = 300,
@@ -176,7 +176,7 @@ class AKSMachineClient(AKSClient):
         with traceback before the exception propagates to ``MachineCRUD``).
 
         Args:
-            agentpool_name: Target agentpool name.
+            agent_pool_name: Target agentpool name.
             vm_size: VM SKU recorded in operation metadata. The PUT body itself
                 only sets ``mode: Machines``; the SKU is informational here so
                 downstream Kusto rows can attribute the agentpool to a SKU.
@@ -194,7 +194,7 @@ class AKSMachineClient(AKSClient):
         cluster_name = cluster_name or self.get_cluster_name()
         metadata = {
             "cluster_name": cluster_name,
-            "agentpool_name": agentpool_name,
+            "agent_pool_name": agent_pool_name,
             "vm_size": vm_size,
         }
         with self._get_operation_context()(
@@ -205,7 +205,7 @@ class AKSMachineClient(AKSClient):
                 url = (
                     f"{_ARM_BASE}/subscriptions/{sub}/resourceGroups/{self.resource_group}"
                     f"/providers/Microsoft.ContainerService/managedClusters/{cluster_name}"
-                    f"/agentPools/{agentpool_name}?api-version={_AGENTPOOL_API_VERSION}"
+                    f"/agentPools/{agent_pool_name}?api-version={_AGENTPOOL_API_VERSION}"
                 )
                 body = {"properties": {"mode": "Machines"}}
                 # Cap the PUT timeout so the bulk of ``timeout`` is preserved
@@ -221,13 +221,13 @@ class AKSMachineClient(AKSClient):
                     )
                 if not self._wait_for_agentpool_provisioning(url, timeout):
                     raise RuntimeError(
-                        f"agentpool {agentpool_name} did not reach Succeeded within {timeout}s"
+                        f"agentpool {agent_pool_name} did not reach Succeeded within {timeout}s"
                     )
                 op.add_metadata("agentpool_info", self.get_node_pool(
-                    agentpool_name, cluster_name).as_dict())
+                    agent_pool_name, cluster_name).as_dict())
                 op.add_metadata("cluster_info", self.get_cluster_data(cluster_name))
             except Exception as e:
-                logger.error(f"Failed to create machine agentpool {agentpool_name}: {e}")
+                logger.error(f"Failed to create machine agentpool {agent_pool_name}: {e}")
                 raise
 
     def _wait_for_agentpool_provisioning(self, url: str, timeout: int) -> bool:
@@ -278,7 +278,7 @@ class AKSMachineClient(AKSClient):
 
     def _wait_for_machine_node_readiness(
         self,
-        agentpool_name: str,
+        agent_pool_name: str,
         expected_count: int,
         timeout: int,
         baseline_count: int = 0,
@@ -314,7 +314,7 @@ class AKSMachineClient(AKSClient):
         }
         if expected_count <= 0:
             return empty_envelope
-        label_selector = f"agentpool={agentpool_name}"
+        label_selector = f"agentpool={agent_pool_name}"
         target_total = baseline_count + expected_count
         # Each percentile target must be STRICTLY GREATER than baseline_count so
         # we count only NEW nodes -- pre-existing labeled Ready nodes alone must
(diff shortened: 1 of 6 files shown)


modules/python/clients/aks_machine_client.py
```diff
@@ -56,6 +55,13 @@
 # ballooning the chunk's wall-time.
 _BATCH_429_MAX_RETRIES = 5
 _BATCH_429_INITIAL_BACKOFF_SECONDS = 1.0
+_BATCH_MAX_MACHINES_PER_REQUEST = 50
+_MACHINE_TERMINAL_STATES = {"Succeeded", "Failed", "Canceled", "Cancelled"}
```
liyu-ma: what is difference between `Canceled` and `Cancelled`

modules/python/clients/aks_machine_client.py
```diff
@@ -383,6 +420,52 @@ def _wait_for_machine_node_readiness(
...
+                    agentpool_name=agentpool_name,
+                    expected_names=expected_machine_names,
+                )
+
+        ready_nodes_result = []
+        failed_machines_result: List[Dict[str, Any]] = []
+        try:
+            ready_nodes_result = ready_nodes.result()
+        except Exception as e:
+            logger.warning(f"get_ready_nodes({label_selector}) failed: {e}")
```
liyu-ma: It is just logging on exception? I think you should throw it rather than swallowing the exception here

modules/python/clients/aks_machine_client.py
```diff
@@ -383,6 +420,52 @@ def _wait_for_machine_node_readiness(
...
+        if failed_machines is not None:
+            try:
+                failed_machines_result = failed_machines.result()
+            except Exception as e:
+                logger.warning(f"get_machine_provisioning_failures failed: {e}")
```
liyu-ma: Same as above. Do you want to always retry on any kind of error?

modules/python/clients/aks_machine_client.py
```diff
@@ -383,6 +420,52 @@ def _wait_for_machine_node_readiness(
         logger.info(f"Percentile node readiness data: {json.dumps(result, indent=2)}")
         return result
 
+    def _get_readiness_poll_results(
+        self,
+        kc: Any,
```
liyu-ma: why do we need to pass `kc`? Shouldn't it part of class's property

modules/python/clients/aks_machine_client.py
liyu-ma: This is not from this PR, but why do we need have `None` check here? `kc` shouldn't be None once it is initialised. We can just check it once in init stage.

modules/python/clients/aks_machine_client.py
```diff
@@ -383,6 +420,52 @@ def _wait_for_machine_node_readiness(
...
+        try:
+            ready_nodes_result = ready_nodes.result()
+        except Exception as e:
+            logger.warning(f"get_ready_nodes({label_selector}) failed: {e}")
+        if failed_machines is not None:
+            try:
+                failed_machines_result = failed_machines.result()
+            except Exception as e:
+                logger.warning(f"get_machine_provisioning_failures failed: {e}")
+            next_machine_failure_check_at = (
```
liyu-ma: When we discover failed machines, why do we still want to keep polling its status next time? Isn't `failed` a terminal state?

modules/python/clients/aks_machine_client.py
```diff
@@ -607,6 +692,91 @@ def _create_single_machine(self, name: str, request: SimpleNamespace) -> bool:
...
+            for machine in machines
+            if machine.get("name") in expected_names
+        }
+        if set(machines_by_name) != expected_names:
+            return []
+        states = {
+            name: machines_by_name[name].get("properties", {}).get("provisioningState")
+            for name in expected_names
+        }
+        if not all(state in _MACHINE_TERMINAL_STATES for state in states.values()):
```
liyu-ma, line 755: Here we return early if some machine is _not_ in terminal state, even if there is mix instances of failed and provisioning. I think it should only return empty array if there is no failed machine.

modules/python/clients/aks_machine_client.py
```diff
@@ -315,17 +324,34 @@ def _wait_for_machine_node_readiness(
         elapsed: Dict[int, float] = {}
```
liyu-ma: The var `elapsed` is not clear - will `completed` be a better name?

modules/python/clients/aks_machine_client.py
```diff
@@ -383,6 +420,52 @@ def _wait_for_machine_node_readiness(
...
+                    expected_names=expected_machine_names,
+                )
+
+        ready_nodes_result = []
+        failed_machines_result: List[Dict[str, Any]] = []
+        try:
+            ready_nodes_result = ready_nodes.result()
+        except Exception as e:
+            logger.warning(f"get_ready_nodes({label_selector}) failed: {e}")
+        if failed_machines is not None:
```
liyu-ma: Shall we do `failed_machines` check before `ready_nodes` check so we can fail fast here?

liyu-ma: Actually, why do we need to call the API twice? The `failed_machines` check list all the machines and you can filter the ready / failed ones. Why not consolidate to just 1 api call over `/machines`

modules/python/clients/aks_machine_client.py
karenychen: thanks for the catch, updated!

modules/python/clients/aks_machine_client.py
karenychen: For Kubernetes node listing, this is intentional: a transient ListNodes failure is retried during the next tick so the percentile readiness loop can keep polling.

iiuc, the current aks client also retries on this failure? https://github.com/Azure/telescope/blob/094218647dd1c13dd6a6f9d719226e8bf3419c30/modules/python/clients/kubernetes_client.py#L240-L262

modules/python/clients/aks_machine_client.py
karenychen: Yes, we retry but only on the existing ListMachines cadence rather than a tight nested retry. ListMachines is heavier at high node counts, so a failure logs and retries on the next scheduled Machine check.

If we get to this stage, it means we have been able to submit all put machine requests and subsequent ListMachines should work without issues. If we are concerned about sustained failures, I can add a retry limit on the list call failures?

modules/python/clients/aks_machine_client.py
karenychen: updated!

modules/python/clients/aks_machine_client.py
karenychen: The parent class's constructor [`AKSClient.__init__`](https://github.com/Azure/telescope/blob/094218647dd1c13dd6a6f9d719226e8bf3419c30/modules/python/clients/aks_client.py#L134-L135) catches Kubernetes client initialization failures and explicitly sets self.k8s_client = None, so readiness code still needs to handle that case.

modules/python/clients/aks_machine_client.py
karenychen: The primary success signal is the node joining the cluster and becoming Ready. Machine Succeeded only means VM/CSE-side provisioning completed from RP's perspective, and it does not 100% guarantee the node joined the cluster. 

ListMachines is only the early-failure signal. Because it is heavier at high counts, we poll it less frequently than k8s list nodes. Our main goal is to prevent the extended timeout duration if we know for sure not all machines will join, otherwise the pipeline might keep polling until timeout and fail.

modules/python/clients/aks_machine_client.py
karenychen: This helper is used as the signal to stop node-readiness polling and fail the operation. If we return failed Machines while the batch is still mixed Failed/Provisioning, we’ll raise immediately and lose P50/P70/P90 readiness data from the Machines that can still join. Once all expected Machines are terminal, no more requested nodes can join, so we return the failed subset and fail the operation then.

modules/python/clients/aks_machine_client.py
karenychen: The failed machine(s) are terminal, but the batch may not be terminal. We are polling further to wait for the remaining expected Machines to reach terminal states. Until then, successful Machines can still join the cluster and contribute lower-percentile readiness data. Once the full expected set is terminal, we return the failed subset and fail the operation.

modules/python/clients/aks_machine_client.py
liyu-ma: Yes parent could set it to None during init phase, then we can just check it once at the end of `AKSMachineClient` init func (and throw ex if needed) rather than every time at call site. I don't see anywhere in the codes this is being set to None. Correct?

modules/python/clients/aks_machine_client.py
```diff
@@ -312,77 +324,153 @@ def _wait_for_machine_node_readiness(
...
+                "percentage": p,
+                "success": p in readiness_times,
+            }
+            for p in (50, 70, 90, 99, 100)
+        }
+
+    def _get_readiness_poll_results(
+        self,
+        label_selector: str,
+        cluster_name: Optional[str],
```
liyu-ma: This function does two things: get ready nodes and machine failure check, which makes the logic tangled and uncomprehensive. The params `cluster_name`, `agentpool_name`, `expected_machine_names` and `next_machine_failure_check_at` are just pass-through here and are all used by machine failure check. I suggest moving the machine failure check logic out of `_get_readiness_poll_results` and into `_wait_for_machine_node_readiness` where you can still use threadpool to invoke them in parallel. That should be a cleaner implementation?

modules/python/clients/aks_machine_client.py
```diff
@@ -312,77 +324,135 @@ def _wait_for_machine_node_readiness(
...
-                logger.warning(f"get_ready_nodes({label_selector}) failed: {e}")
-                ready = 0
+        while time.time() < deadline and len(readiness_times) < len(targets):
+            failed_machines = None
+            should_check_machines = (
+                next_machine_failure_check_at is not None
+                and time.time() >= next_machine_failure_check_at
+            )
+            max_workers = 2 if should_check_machines else 1
+            with ThreadPoolExecutor(max_workers=max_workers) as ex:
```
liyu-ma: If there is only one task, a threadpool for each wait cycle is wasteful and overhead. Lets do some optimsation like this:
```
if should_check_machines and cluster_name and expected_machine_names:
    with ThreadPoolExecutor(max_workers=2) as ex:
        ready_fut = ex.submit(self._get_ready_node_count, label_selector)
        failed_fut = ex.submit(self._get_terminal_machine_provisioning_failures, ...)
        ready = ready_fut.result()
        try:
            failed_machines_result = failed_fut.result()
        except Exception as e:
            ...
else:
    ready = self._get_ready_node_count(label_selector)
    failed_machines_result = []

```

modules/python/clients/aks_machine_client.py
```diff
@@ -56,6 +55,13 @@
 # ballooning the chunk's wall-time.
 _BATCH_429_MAX_RETRIES = 5
 _BATCH_429_INITIAL_BACKOFF_SECONDS = 1.0
+_BATCH_MAX_MACHINES_PER_REQUEST = 50
+_MACHINE_TERMINAL_STATES = {"Succeeded", "Failed", "Canceled"}
+_MACHINE_FAILURE_STATES = {"Failed", "Canceled"}
+_MACHINE_FAILURE_DETAIL_LIMIT = 10
+_LIST_MACHINES_PAGE_LIMIT = 50
```
liyu-ma: The naming is misleading. I think `_LIST_MACHINES_MAX_PAGES` is more accurate

modules/python/clients/aks_machine_client.py
```diff
@@ -64,11 +70,27 @@
...
+# (50 individual workers, 10 batch workers) with comfortable headroom and
 # stays well below ARM's per-client connection ceilings.
 _HTTPS_POOL_SIZE = 64
 
 
+class MachineProvisioningFailed(RuntimeError):
+    """Terminal Machine failure with readiness metadata captured before failure."""
+    def __init__(
+        self,
+        agentpool_name: str,
```
liyu-ma: Nit: "agent pool" is two-words. Can we replace `agentpool_name` with `agent_pool_name` everywhere?

liyu-ma (approved):

PR #1213, opened 2026-06-08
[fix] Retry on OperationNotAllowed when adding node pools

liyu-ma:
Make the `nodepool add` operation more resilient by adding retry mechanism.

### DIFF
--- a/kcl/lib/steps/azure/create_node_pool.k
+++ b/kcl/lib/steps/azure/create_node_pool.k
@@ -25,7 +25,7 @@ CreateNodePool = lambda serviceConnection: str, cluster: str, resourceGroup: str
         "--no-wait" if noWait else "",
     ]
     flags = " \\\n  ".join([f for f in flagDecision if f])
-    # Allow retry on OperationNotAllowed — it's observed that AKS can spin up operation in the background that causes transient OperationNotAllowed errors on node pool PUT requests
+    # Allow retry on OperationNotAllowed — it's observed that AKS can spin up PutExtensionAddonHandler operation in the background that causes transient OperationNotAllowed errors on node pool PUT requests
     script = """
 deadline=$(($(date +%s) + ${CREATE_NODEPOOL_RETRY_WINDOW_SEC}))
 sleep_between=30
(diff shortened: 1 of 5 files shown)


kcl/lib/steps/azure/create_node_pool.k
```diff
@@ -22,16 +25,42 @@ CreateNodePool = lambda serviceConnection: str, cluster: str, resourceGroup: str
         "--no-wait" if noWait else "",
     ]
     flags = " \\\n  ".join([f for f in flagDecision if f])
+    # Allow retry on OperationNotAllowed — it's observed that AKS can spin up operation in the background that causes transient OperationNotAllowed errors on node pool PUT requests
```
wonderyl: What operations are they?

kcl/lib/steps/azure/create_node_pool.k
```diff
@@ -22,16 +25,42 @@ CreateNodePool = lambda serviceConnection: str, cluster: str, resourceGroup: str
...
+  rc=$?
+  set -e
+  echo "$output"
+  if [ $rc -eq 0 ]; then
+    exit 0
+  fi
+  if ! echo "$output" | grep -qi "OperationNotAllowed"; then
+    echo "Non-retriable error (rc=$rc); giving up after attempt $attempt."
+    exit $rc
+  fi
```
wonderyl: nit: imagine when you have another type of output, your whole code would need a big restructure. This is because you use `if` to scope in a more generic class of condition, and leave a more specific class of condition, i.e. OperationNotAllowed, as the default behavior.

kcl/lib/steps/azure/create_node_pool.k
liyu-ma: It was `PutExtensionAddonHandler` request. I will update comment here

kcl/lib/steps/azure/create_node_pool.k
```diff
@@ -22,16 +25,42 @@ CreateNodePool = lambda serviceConnection: str, cluster: str, resourceGroup: str
...
+    ${flags} 2>&1)
+  rc=$?
+  set -e
+  echo "$output"
+  if [ $rc -eq 0 ]; then
+    exit 0
+  fi
+  if ! echo "$output" | grep -qi "OperationNotAllowed"; then
+    echo "Non-retriable error (rc=$rc); giving up after attempt $attempt."
+    exit $rc
```
xinWeiWei24: I also ran into this issue before — it said another operation was already running, but after waiting around 5 minutes, I was still able to create the nodepool. Is this the same issue?

liyu-ma: yes it is. It is transient so retry will pass evetually.

kcl/lib/steps/azure/create_node_pool.k
liyu-ma: Now using `case` statement.

wonderyl (approved):

PR #1221, opened 2026-06-16
Add step names and outputs for vnet and identity steps

xinWeiWei24:
- Add explicit step names (createSubnet, createManagedIdentity) and isOutput flag for clear cross-step variable references to improves pipeline traceability and code readability by making variable sources explicit, e.g. using createManagedIdentity.IDENTITY_ID for accessing the exported value.
- VNet-level role assignments enable subnets to inherit permissions from parent VNet


kcl/lib/steps/azure/az_cli.k
```diff
@@ -1,6 +1,6 @@
 import azure_pipelines.ap.steps
 
-AzCli = lambda serviceConnection: str, name: str, script: str, condition: str = Undefined, continueOnError: bool = Undefined -> steps.Step {
+AzCli = lambda serviceConnection: str, name: str, script: str, condition: str = Undefined, continueOnError: bool = Undefined, stepName: str = Undefined -> steps.Step {
```
liyu-ma: can we also rename `name` to `displayName`

kcl/lib/steps/azure/vnet.k
```diff
@@ -1,17 +1,31 @@
 import azure_pipelines.ap.steps
 
-CreateVNet = lambda serviceConnection: str, resourceGroup: str, name: str, addressPrefixes: str, subscription: str -> steps.Step {
+CreateVNet = lambda serviceConnection: str, resourceGroup: str, name: str, addressPrefixes: str, subscription: str, assignee: str = "" -> steps.Step {
+    roleAssignmentScript = """
```
liyu-ma: no need to have `roleAssignmentScript` var, you can just inline it within
```
azure.AzCli(
 ...
)
```

kcl/lib/steps/azure/identity.k
liyu-ma: Same here, you can just inline `script` within `AzCli( )`

kcl/lib/steps/azure/az_cli.k
```diff
@@ -12,5 +12,6 @@ AzCli = lambda serviceConnection: str, name: str, script: str, condition: str =
         displayName = name
         condition = condition
         continueOnError = continueOnError
+        name = stepName
```
wonderyl, line 15: Do you need display name different than name? why not set it to name?

xinWeiWei24: They serve different roles. name (→ displayName) is the human-readable UI label and can be verbose. stepName (→ name) is the step identifier used in pipeline expressions like $(steps.<stepName>.outputs.<var>), so it needs to be short and follow variable-name conventions. Keeping them separate lets callers set a descriptive display label while still choosing a concise identifier for output references.

kcl/lib/steps/azure/vnet.k
```diff
@@ -1,17 +1,29 @@
...
+  --role "Network Contributor" \\
+  --assignee-object-id "${assignee}" \\
+  --assignee-principal-type ServicePrincipal \\
+  --subscription "${subscription}"
+""" if assignee else ""))
 }
 
-CreateSubnetAndRoleAssignment = lambda serviceConnection: str, resourceGroup: str, vnetName: str, name: str, addressPrefixes: str, subscription: str, assignee: str -> steps.Step {
+CreateSubnet = lambda serviceConnection: str, resourceGroup: str, vnetName: str, name: str, addressPrefixes: str, subscription: str -> steps.Step {
     script = """
```
liyu-ma: nitpick: no need to have the `script` var, can you also inline it within `AzCli` call

liyu-ma (approved):

kcl/lib/steps/azure/az_cli.k
wonderyl: Thanks for explaining. I am fine with this pr. I was thinking whether we can use this output feature to make the code better. One of the problem is the generated env cars are like backdoor and it not explicitly declared anywhere except for in the code. We can even verify a step generating n env var is included in a job or not. We don't need to resolve it right away, but something to think about. Feel free to resolve this comment after reading.

wonderyl (approved):

kcl/lib/steps/azure/az_cli.k
xinWeiWei24: Yes, I think we’re aligned on the main issue: environment variables are not easily traceable and reduce readability. Since we don’t currently have a “module” concept, I changed the implementation so each step function can expose outputs, allowing other steps to reference them through the function name that generated the value.

PR #1228, opened 2026-06-19
Add reusable jumpbox step under lib/steps/azure

xinWeiWei24:
Provide a generic azure.CreateJumpbox(...) lambda so any pipeline can provision a Linux jumpbox VM with a managed identity and AKS role assignments.

### DIFF
--- a/kcl/lib/steps/azure/jumpbox.k
+++ b/kcl/lib/steps/azure/jumpbox.k
@@ -38,11 +38,12 @@ DEFAULT_AKS_JUMPBOX_ROLES = [
     RoleAssignment {role = "Azure Kubernetes Service Contributor Role"}
 ]
 
-# CreateJumpbox provisions a Linux jumpbox VM in an existing VNet/subnet, opens SSH, enables a system-assigned
+# CreateJumpbox provisions a Linux jumpbox VM in an existing VNet/subnet, enables a system-assigned
 # managed identity, and (optionally) assigns roles to that identity.
 #
-# Authentication uses `az vm create --generate-ssh-keys`, which auto-generates an SSH key pair at
-# ~/.ssh/id_rsa{,.pub} on the pipeline agent (ephemeral, lost when the agent terminates).
+# Remote access is expected to use `az vm run-command invoke` (ARM -> VM Guest Agent), so no inbound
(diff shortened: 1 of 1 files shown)


kcl/lib/steps/azure/jumpbox.k
```diff
@@ -0,0 +1,98 @@
...
+# must wire that variable up via a variable group or pipeline secret.
+CreateJumpbox = lambda \
+    serviceConnection: str, \
+    subscription: str, \
+    resourceGroup: str, \
+    name: str, \
+    vnetName: str, \
+    subnetName: str = "jumpbox", \
+    adminUsername: str = "azureuser", \
+    adminPasswordVar: str = "JUMPBOX_ADMIN_PASSWORD", \
```
liyu-ma: do we have to user username and password? Can we use ssh key like what we did here https://github.com/Azure/telescope/blob/main/steps/ssh/setup-key.yml

wonderyl: +1, though not necessarily ssh key, but randomly generate the password and store into key vault makes it more secure.

kcl/lib/steps/azure/jumpbox.k
```diff
@@ -0,0 +1,98 @@
...
+    vmSize: str = "Standard_D16ds_v5", \
+    image: str = "Ubuntu2204", \
+    cloudInit: str = DEFAULT_JUMPBOX_CLOUD_INIT, \
+    roles: {str:str} = DEFAULT_AKS_JUMPBOX_ROLES \
```
wonderyl: use a schema instead of dict, otherwise it's pretty hard to understand what the key and value represent without reading the code.

kcl/lib/steps/azure/jumpbox.k
xinWeiWei24: Switched to az vm create --generate-ssh-keys. Subsequent automation runs on the same agent, so it can use the agent's local key directly

kcl/lib/steps/azure/jumpbox.k
xinWeiWei24: done

wonderyl (approved):

PR #1231, opened 2026-06-23
Register crud module in pyproject.toml

diamondpowell:
## Summary

Registers the open-source CRUD module as part of the `telescope-benchmarks` Python package so callers can install and run CRUD operations without source-tree `PYTHONPATH` routing.

## Why

The internal Telescope CRUD migration now uses the open-source Python CRUD module as the source of truth for node-pool and workload operations. This PR provides the package contract required by that migration: installable modules, runtime workload manifests, a CLI entry point, and explicit pipeline authentication and cluster selection inputs.

## Changes

- Register `crud`, `crud.azure`, and `crud.aws` in `pyproject.toml`.
- Add the `telescope-crud` console entry point.
- Include deployment, Job, and StatefulSet YAML manifests in the installed package.
- Consume the credential contract merged in [PR #1305](https://github.com/Azure/telescope/pull/1305): shared `configure_credential` handling and explicit credential injection into `AKSClient`.
- Keep Azure credential creation in the CRUD callers and expose managed-identity exclusion as an opt-in CLI flag.
- Add optional AKS cluster-name selection while preserving discovery when omitted.
- Preserve the existing positional `NodePoolCRUD` constructor contract.
- Add focused tests for authentication behavior and constructor compatibility.

## Related PRs

Credential refactor: [PR #1305](https://github.com/Azure/telescope/pull/1305) merged on September 2, 2026 at merge commit `6b9fbaad7ae092d7b96fe7451f6b7987de314d22`. This PR is rebased onto that contract at `d41f41bf3536a0c1c4145c3aea127c4d67e10c61`.

Internal migration: [ADO PR 16193735](https://msazure.visualstudio.com/CloudNativeCompute/_git/telescope/pullrequest/16193735)

The two PRs are separated by ownership boundary:

- This PR owns the reusable open-source Python package and runtime contract.
- The internal PR owns pipeline orchestration, cloud login, resource validation, result publication, and production rollout.

The internal PR is intentionally open as this package's integration consumer. It shows reviewers how the package is installed and invoked, so review of both PRs can happen in parallel. The internal PR does not need to merge before this package PR.

The internal validation pipeline cloned this draft branch directly and installed `modules/python` with pip. Draft status does not affect branch installation as long as the branch is pushed.

## Validation

- Final open-source head: `d41f41bf3536a0c1c4145c3aea127c4d67e10c61`.
- 797 tests passed, 1 skipped; coverage 82.99%; Pylint 10.00/10.
- GitHub Python Validation, YAML Validation, post-comment, and CLA checks passed.

- 100 focused CRUD and AKS client unit tests passed before pipeline validation.
- 81 focused CRUD tests passed after the constructor-compatibility fix.
- Python Validation: passed.
- YAML Validation: passed.
- CLA: passed.
- Feature-to-feature pip-install validation: [ADO build 177409535](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177409535).
  - Five of five Azure regions passed: `westcentralus`, `westus2`, `eastasia`, `eastus2euap`, and `centraluseuap`.
  - No failed or canceled records.
  - Internal commit: `c18584d6`.
  - Open-source package commit: `3e335184`.

### Post-Refactor `new-pipeline-test.yml` Validation

- [ADO build 179423290](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=179423290) passed using the temporary `new-pipeline-test.yml` configuration at internal test commit `b9460721`.
- The run installed and validated open-source package commit `d41f41bf`, which includes the merged PR #1305 credential refactor.
- Cleanup commit `3730f654` restored `new-pipeline-test.yml` after validation; the temporary pipeline YAML is not part of this open-source PR.

## Review and Merge Sequence

1. Review this package PR and internal PR #16193735 in parallel.
2. Merge this package PR first. Package registration remains separate and is not blocked on workload timing.
3. Review and merge the planned workload-timing PR against the updated open-source `main`.
4. Revalidate the already-reviewed internal migration against the resulting open-source `main` from a disposable validation branch.
5. Merge internal PR #16193735 last, after required workload telemetry and consumer compatibility pass.
6. Run a post-merge production smoke test and validate ADX output.

## Final Follow-up Validation

- Earlier open-source commit: `9f96c3e0`; superseded by post-refactor validation at `d41f41bf`.
- 123 focused tests passed after the VirtualMachines scale-state, exit-code, and helper/test extraction changes.
- Pylint: `10.00/10`.
- Final file sizes: `aks_client.py` 974 lines, `test_aks_client.py` 838 lines, and `test_main.py` 1564 lines.
- Targeted VirtualMachines lifecycle: [ADO build 177546874](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177546874).
  - Internal test commit: `00c99b90`.
  - Open-source package commit: `9f96c3e0`.
  - Create, scale up, Deployment, StatefulSet, Job, scale down, delete, collect, upload, and cleanup all passed.
  - No failed, canceled, or succeeded-with-issues records.

Build `177409535` remains the five-region VMSS installation validation; build `177546874` is the targeted VirtualMachines validation.

### Earlier Authoritative Multi-Region Validation

- **Earlier Build:** [ADO build 177701638](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177701638) (Aug 21, 2026)
- **Commits:** Internal test commit `b9460721` and open-source tested commit `9f96c3e0`. Internal cleanup commit `3730f654` restored `new-pipeline-test.yml` (cleanup commit itself was not tested by build).
- **Regions:** `westcentralus`, `westus2`, `eastasia`, `eastus2euap`, and `centraluseuap`. Both small VMSS and VirtualMachines cases ran in every region; 5/5 stages and 10/10 matrix jobs passed.
- **Scope:** Package install, create, scale up, Deployment, StatefulSet, Job, scale down, delete, collect/upload/cleanup passed.
- **Status:** No failed, canceled, abandoned, or succeeded-with-issues records.
- **Distinction:** Clearly retain build `177409535` as earlier five-region VMSS package-installation evidence and build `177546874` as earlier one-region targeted VirtualMachines evidence; do not call either final-head all-type regional evidence.


modules/python/crud/main.py
```diff
@@ -449,6 +454,11 @@ def main():
         default="ON_DEMAND",
         help="Capacity type for AWS/Azure node pool",
     )
+    common_parser.add_argument(
+        "--exclude-managed-identity",
+        action="store_true",
+        help="Exclude managed identity from DefaultAzureCredential",
+    )
```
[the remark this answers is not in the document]
diamondpowell: Good point. I scoped these options to node-pool and workload commands, so machine commands no longer accept them. Non-Azure node-pool and workload commands now reject the Azure-only options explicitly. I also added tests for Azure forwarding, machine-command rejection, and AWS rejection in b1e0526a.

modules/python/utils/provisioning_instrumentation.py
```diff
@@ -83,6 +83,47 @@ def instrument_nodepool_provisioning(
     return ready_nodes
 
 
+def instrument_aks_nodepool_provisioning(
```
liyu-ma: I don't see why we need this extra wrapper. The name is so close to `instrument_nodepool_provisioning` and this makes people confused. I don't see how this help generalize our codes.

modules/python/utils/azure_node_pool_cli.py
```diff
@@ -0,0 +1,239 @@
...
+    cluster_name,
+    node_pool_name,
+    node_count,
+    vm_size,
+):
+    """Configure VMSS parameters or return a VirtualMachines create callable."""
+    if node_pool_type != "VirtualMachines":
+        parameters["count"] = node_count
+        parameters["vm_size"] = vm_size
+        return None
```
liyu-ma: Here is the code smell:
Line 144 returns None but also mutate the parameters _secretly_, while line 147 returns a function. Check the call site - the None value of arm_callable is passed through but eventually fallback to `begin_create_or_update_with_retry` within provisioning_instrumentation.py.

How about something like this:
```
def prepare_create_operation(
    parameters, node_pool_type, gpu_node_pool,
    resource_group, cluster_name, node_pool_name, node_count, vm_size,
    *, aks_client, k8s_client=None,  # whatever the SDK callable needs
):
    """Return a zero-arg ARM callable for the create, dispatched by pool type."""
    if node_pool_type == "VirtualMachines":
        if gpu_node_pool:
            raise ValueError("GPU node pools with type VirtualMachines are not supported")
        return partial(
            create_virtual_machines_node_pool,
            resource_group, cluster_name, node_pool_name, node_count, vm_size,
        )
    # VMSS: still the SDK path, but returned as a callable instead of a None+mutation
    parameters = {**parameters, "count": node_count, "vm_size": vm_size}
    return partial(begin_create_or_update_with_retry,
                   aks_client, resource_group, cluster_name, node_pool_name, parameters)
```
to make it always return a callable.

The same reasoning can apply to `def prepare_scale_operation` below

modules/python/clients/aks_client.py
```diff
@@ -428,6 +400,7 @@ def create_node_pool(
         enable_managed_gpu: bool = False,
         gpu_instance_profile: Optional[str] = None,
         gpu_mig_strategy: Optional[str] = None,
+        node_pool_type: str = "VirtualMachineScaleSets",
```
liyu-ma: Can we make `VirtualMachineScaleSets` as constant? I see multiple places where we have this hard-coded string.

modules/python/utils/azure_node_pool_cli.py
```diff
@@ -0,0 +1,239 @@
...
+from functools import partial
+
+from utils.logger_config import get_logger
+
+logger = get_logger(__name__)
+
+
+def get_node_pool_scale_state(node_pool, node_pool_type):
+    """Return current node count and VM size for VMSS or VirtualMachines pools."""
+    if node_pool_type != "VirtualMachines":
```
liyu-ma: Can we make `VirtualMachines` as constant? Can you inspect all occurrences of this hard-coded str from this PR and update them

modules/python/crud/main.py
```diff
@@ -126,8 +126,10 @@ def handle_node_pool_operation(node_pool_crud, args):
                 "node_count": args.node_count,
                 "gpu_node_pool": args.gpu_node_pool,
                 "enable_managed_gpu": args.enable_managed_gpu,
-                **azure_gpu_kwargs,
+                **azure_kwargs,
             }
+            if args.cloud == "azure":
+                create_kwargs["node_pool_type"] = args.node_pool_type
```
liyu-ma: There are many occurences like line131-132. You can just fold this to 
```
azure_kwargs = {
    "gpu_instance_profile": args.gpu_instance_profile,
    "gpu_mig_strategy": args.gpu_mig_strategy,
    "node_pool_type": args.node_pool_type,
}
```

modules/python/utils/provisioning_instrumentation.py
diamondpowell: I added the wrapper when I added VirtualMachines support because the new path uses an Azure CLI callable while the existing VMSS path uses the SDK. I needed both operations to use the same ARM and Kubernetes readiness instrumentation.

Looking at the resulting call chain, I agree that the wrapper does not provide enough separation to justify another similarly named function. I removed it. AKSClient now builds the Kubernetes readiness callable and passes it, along with the prepared ARM callable, directly to instrument_nodepool_provisioning.

modules/python/utils/azure_node_pool_cli.py
diamondpowell: I agree that returning either None or a callable makes the contract difficult to follow. I originally used None to preserve the existing VMSS fallback while I added the VirtualMachines CLI path. I did not realize until your comment that this split the final operation selection between the preparation helper and the instrumentation wrapper.

I updated both prepare_create_operation and prepare_scale_operation to always return a zero-argument ARM callable. VMSS now returns a callable around begin_create_or_update_with_retry, while VirtualMachines returns the Azure CLI callable.

I also moved the GPU profile setup before the VMSS parameter copy and preserved the progressive-scale label. The tests now execute the VMSS callables and verify the SDK payloads, including progressive counts [2, 3].

modules/python/clients/aks_client.py
diamondpowell: I agree that this value should not be repeated across the different layers. I added AzureNodePoolTypeConstants.VIRTUAL_MACHINE_SCALE_SETS and used it for the production defaults and comparisons in AKSClient, Azure NodePoolCRUD, and the CLI dispatch.

modules/python/utils/azure_node_pool_cli.py
diamondpowell: I checked all occurrences of the hard-coded VirtualMachines value in my changes. I replaced the production comparisons, defaults, and Azure command construction with AzureNodePoolTypeConstants.VIRTUAL_MACHINES.

I kept the literal in error messages, docstrings, and the test that verifies the exact Azure CLI command because those are user-facing text or external-contract assertions.

modules/python/crud/main.py
diamondpowell: I agree that node_pool_type belongs with the other Azure-only arguments. I added it to the shared azure_kwargs dictionary and removed the repeated assignments from the create, scale, and all branches.

I scoped the dictionary to those three commands because delete uses the same handler but does not define node_pool_type, gpu_instance_profile, or gpu_mig_strategy. This keeps the shared construction without breaking Azure delete or forwarding Azure-only arguments to AWS.

modules/python/clients/aks_client.py
```diff
@@ -120,6 +116,11 @@ def __init__(
                           will try to get the first cluster in the resource group.
             use_managed_identity: Whether to use managed identity for authentication.
                                  If False, will fall back to DefaultAzureCredential.
+            exclude_managed_identity: When using DefaultAzureCredential, exclude the
```
liyu-ma: I did not catch this change previously but this is making things very complicated. Note that this func also takes a `use_managed_identity` but it is for different purpose. Since you are touching this part, I would request you to make **a separate PR** to refactor the construction of credential for aks_client here.

Here is my proposal:
- Create a new file `azure_auth.py` under `/utils` folder
- Have a new func `def configure_credential(use_managed_identity=False, default_credential_exclude_mi=False)` that capture the credential creation logic here and returns a credential object
- Update client sites to pass in created credential when creating aks client, and removing two flags from this function. 
- Add/update all tests

diamondpowell: I have created this pr and added tests and ran the pipeline as well
https://github.com/Azure/telescope/pull/1305/

diamondpowell: merged and pr rebased to include auth refactor

modules/python/utils/azure_node_pool_cli.py
```diff
@@ -0,0 +1,264 @@
...
+        resource_group,
+        "--cluster-name",
+        cluster_name,
+        "--name",
+        node_pool_name,
+        "--node-count",
+        str(node_count),
+        "--node-vm-size",
+        vm_size,
+        "--vm-set-type",
```
liyu-ma: Why do we need to add `--vm-set-type` for GPU pool? This is a behavior change different from current implementation: 
https://github.com/Azure/telescope/blob/main/modules/python/clients/aks_client.py#L289-L320
Will this break things? @xuexu6666

diamondpowell: Ahhh it isn’t needed for the managed GPU path. I accidentally passed it through as well while adding VirtualMachines support to the non-GPU path. I removed it from the managed GPU command and added regression coverage to keep the paths separate.

liyu-ma (approved):

PR #1234, opened 2026-06-25
feat(kcl/k8s): add ValidateWorkloadHealth step

xinWeiWei24:
Generic K8s workload health-check step for Azure Pipelines. Supports Deployment, StatefulSet, DaemonSet, Job, Pod, and wildcard ("*") on a namespace. Verifies controller readiness / Job completion / Pod phase, then scans container logs for panic|fatal (overridable).

### DIFF
--- a/kcl/lib/steps/k8s/validate_workload.k
+++ b/kcl/lib/steps/k8s/validate_workload.k
@@ -64,54 +64,50 @@ check_controllers() {
   done
 }
 
-check_pod() {
-  local pod=$1
-  echo "::: checking pod ${namespace}/$pod"
-  local phase ready_cond not_ready log_output matches
-  phase=$(kubectl -n "${namespace}" get pod "$pod" -o jsonpath='{.status.phase}' 2>/dev/null)
-  case "$phase" in
-    Succeeded)
-      # Job pod that finished cleanly; skip readiness checks, still scan logs.
-      ;;
-    Running)
-      ready_cond=$(kubectl -n "${namespace}" get pod "$pod" \\
-        -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}')
-      if [ "$ready_cond" != "True" ]; then
-        fail "pod ${namespace}/$pod Ready condition=$ready_cond (expected True)"
-        return
-      fi
-      not_ready=$(kubectl -n "${namespace}" get pod "$pod" \\
-        -o jsonpath='{range .status.containerStatuses[?(@.ready==false)]}{.name} {end}')
-      if [ -n "$not_ready" ]; then
-        fail "pod ${namespace}/$pod has not-ready containers: $not_ready"
-        return
-      fi
-      ;;
-    *)
-      fail "pod ${namespace}/$pod phase=$phase (expected Running or Succeeded)"
-      return
-      ;;
-  esac
-  log_output=$(kubectl -n "${namespace}" logs "$pod" --all-containers=true --prefix=true 2>&1)
-  matches=$(printf '%s\\n' "$log_output" | grep -wEi '${errorPatterns}')
-  if [ -n "$matches" ]; then
-    echo "--- error lines in ${namespace}/$pod ---"
-    printf '%s\\n' "$matches"
-    echo "--- end ---"
-    fail "pod ${namespace}/$pod logs contain error pattern"
-  fi
-}
-
 # check_pods POD_LIST
-# Scan each pod in a whitespace-separated list. Fails if the list is empty.
+# Check each pod in a whitespace-separated list: status is Running (Ready
+# condition True, all containers ready) or Succeeded, then scan its logs for
+# errorPatterns. Fails if the list is empty.
 check_pods() {
-  local pods=$1 pod
+  local pods=$1 pod phase ready_cond not_ready log_output matches
   if [ -z "$pods" ]; then
     fail "no pods found for ${desc}"
     return
   fi
   for pod in $pods; do
-    check_pod "$pod"
+    echo "::: checking pod ${namespace}/$pod"
+    phase=$(kubectl -n "${namespace}" get pod "$pod" -o jsonpath='{.status.phase}' 2>/dev/null)
+    case "$phase" in
+      Succeeded)
+        # Finished cleanly; skip readiness checks, still scan logs below.
+        ;;
+      Running)
+        ready_cond=$(kubectl -n "${namespace}" get pod "$pod" \\
+          -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}')
+        if [ "$ready_cond" != "True" ]; then
+          fail "pod ${namespace}/$pod Ready condition=$ready_cond (expected True)"
+          continue
+        fi
+        not_ready=$(kubectl -n "${namespace}" get pod "$pod" \\
+          -o jsonpath='{range .status.containerStatuses[?(@.ready==false)]}{.name} {end}')
+        if [ -n "$not_ready" ]; then
+          fail "pod ${namespace}/$pod has not-ready containers: $not_ready"
+          continue
+        fi
+        ;;
+      *)
+        fail "pod ${namespace}/$pod phase=$phase (expected Running or Succeeded)"
+        continue
+        ;;
+    esac
+    log_output=$(kubectl -n "${namespace}" logs "$pod" --all-containers=true --prefix=true 2>&1)
+    matches=$(printf '%s\\n' "$log_output" | grep -wEi '${errorPatterns}')
+    if [ -n "$matches" ]; then
+      echo "--- error lines in ${namespace}/$pod ---"
+      printf '%s\\n' "$matches"
+      echo "--- end ---"
+      fail "pod ${namespace}/$pod logs contain error pattern"
+    fi
   done
 }
 
--- a/kcl/lib/steps/k8s/validate_workload.k
+++ b/kcl/lib/steps/k8s/validate_workload.k
@@ -34,7 +34,7 @@ ValidateWorkloadHealth = lambda \
 set +e
 
 failures=0
-fail() { echo "FAIL: $*"; failures=$((failures + 1)); }
+fail() { echo "FAILED WORKLOAD CHECK: $*"; failures=$((failures + 1)); }
 
 # check_controllers KIND DESIRED_PATH READY_PATH [NAME]
 # If NAME is given, check just that one; otherwise check every instance of KIND.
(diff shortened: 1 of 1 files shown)


kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -0,0 +1,181 @@
...
+ValidateWorkloadHealth = lambda \
+    serviceConnection: str, \
+    kind: str, \
+    name: str = "", \
+    namespace: str = "default", \
+    errorPatterns: str = "panic|fatal", \
+    displayName: str = "", \
+    continueOnError: bool = Undefined \
+    -> steps.Step {
+    assert kind in ["deployment", "statefulset", "daemonset", "job", "pod", "*"], \
```
wonderyl: use can define kind as 
```
kind: "deployment" | "statefulset" | "daemonset" | "job" | "pod"| "*"
```

kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -0,0 +1,181 @@
...
+    desc = "all workloads in ${namespace}" if kind == "*" else "${kind} ${namespace}/${name}"
+    script = """
+set +e
+
+failures=0
+fail() { echo "FAIL: $*"; failures=$((failures + 1)); }
+
+# check_controllers KIND DESIRED_PATH READY_PATH [NAME]
+# If NAME is given, check just that one; otherwise check every instance of KIND.
+check_controllers() {
```
wonderyl, line 41: does this only check controllers, not any workload?

xinWeiWei24: check_controllers is one of three helpers (check_controllers / check_jobs / check_pod). They are split because the K8s status fields differ:

Deployment / StatefulSet / DaemonSet → replicas vs ready (handled here)
Job → completions / succeeded / active / Failed condition (→ check_jobs)
Pod → phase + Ready condition + container readiness (→ check_pod)
The top-level case "${kind}" dispatches to the right one, so the step does cover all workload kinds

kcl/lib/steps/k8s/validate_workload.k
xinWeiWei24: Done!

kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -0,0 +1,179 @@
...
+  if [ -n "$matches" ]; then
+    echo "--- error lines in ${namespace}/$pod ---"
+    printf '%s\\n' "$matches"
+    echo "--- end ---"
+    fail "pod ${namespace}/$pod logs contain error pattern"
+  fi
+}
+
+# 1. Resolve target controllers and pod set.
+case "${kind}" in
```
wonderyl, line 130: It's clearer to write the following way. You used double case, to reuse the same piece of code, but you can achieve the same with a function.
```
case "${kind}" in
  "*")
    check_controllers deployment  '{.spec.replicas}'                 '{.status.readyReplicas}'
    check_controllers statefulset '{.spec.replicas}'                 '{.status.readyReplicas}'
    check_controllers daemonset   '{.status.desiredNumberScheduled}' '{.status.numberReady}'
    check_jobs ""
    pods=$(kubectl -n "${namespace}" get pods -o jsonpath='{.items[*].metadata.name}')
    ;;
  pod)
    check_pods "${name}"
    ;;
  deployment|statefulset)
    check_controllers "${kind}" '{.spec.replicas}' '{.status.readyReplicas}' "${name}" 
    check_controller_pods "${kind}" "${name}"
    ;;
  daemonset)
    check_controllers daemonset '{.status.desiredNumberScheduled}' '{.status.numberReady}' "${name}" ;;
    check_controller_pods "${kind}" "${name}"
  job)
    check_jobs "${name}" 
    check_controller_pods "${kind}" "${name}"
    ;;
esac
```

kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -0,0 +1,179 @@
...
+        "errorPatterns must not contain single quotes (got ${errorPatterns})"
+
+    defaultTitle = "Validate all workloads in ${namespace}" if kind == "*" else "Validate ${kind}/${name}"
+    title = displayName if displayName else defaultTitle
+    desc = "all workloads in ${namespace}" if kind == "*" else "${kind} ${namespace}/${name}"
+    script = """
+set +e
+
+failures=0
+fail() { echo "FAIL: $*"; failures=$((failures + 1)); }
```
wonderyl: When you read log and try to find the exact failure, you will search for "FAIL:", but this string to too simple, other error string may have the same pattern. I suggest to add a bit more words to make it less likely to collide with others. E.g. "FAILED WORKLOAD CHECK:"

kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -0,0 +1,179 @@
...
+  else
+    names=$(kubectl -n "${namespace}" get jobs -o jsonpath='{.items[*].metadata.name}')
+  fi
+  for n in $names; do
+    completions=$(kubectl -n "${namespace}" get job "$n" -o jsonpath='{.spec.completions}')
+    succeeded=$(kubectl -n "${namespace}" get job "$n" -o jsonpath='{.status.succeeded}')
+    active=$(kubectl -n "${namespace}" get job "$n" -o jsonpath='{.status.active}')
+    failed_cond=$(kubectl -n "${namespace}" get job "$n" \\
+      -o jsonpath='{.status.conditions[?(@.type=="Failed")].status}')
+    [ -z "$completions" ] && completions=1
```
wonderyl: avoids silence failure, not being able to get comletions and other fields should be considered as failures.

kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -0,0 +1,179 @@
...
+    [ -z "$succeeded" ] && succeeded=0
+    [ -z "$active" ] && active=0
+    echo "job ${namespace}/$n: succeeded=$succeeded/$completions active=$active failed=$failed_cond"
+    if [ "$failed_cond" = "True" ]; then
+      fail "job ${namespace}/$n failed"
+    elif [ "$succeeded" -lt "$completions" ] && [ "$active" = "0" ]; then
```
wonderyl: So if I want 3 completions, 2 succeeded and 1 is active, it will count as succeed?

kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -0,0 +1,179 @@
...
+    if [ -z "$selector" ]; then
+      fail "${kind} ${namespace}/${name} has no matchLabels selector (matchExpressions not supported)"
+      pods=""
+    else
+      pods=$(kubectl -n "${namespace}" get pods -l "$selector" -o jsonpath='{.items[*].metadata.name}')
+    fi
+    ;;
+esac
+
+if [ -z "$pods" ]; then
```
wonderyl: $pods is assigned in multiple places and used in here. Having a local variable that span 10s of lines of code makes it harder to read.  If you use the way I suggested in https://github.com/Azure/telescope/pull/1234/changes#r3488891510, you can avoid this.

kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -0,0 +1,179 @@
...
+# 1. Resolve target controllers and pod set.
+case "${kind}" in
+  "*")
+    check_controllers deployment  '{.spec.replicas}'                 '{.status.readyReplicas}'
+    check_controllers statefulset '{.spec.replicas}'                 '{.status.readyReplicas}'
+    check_controllers daemonset   '{.status.desiredNumberScheduled}' '{.status.numberReady}'
+    check_jobs ""
+    pods=$(kubectl -n "${namespace}" get pods -o jsonpath='{.items[*].metadata.name}')
+    ;;
+  pod)
```
wonderyl: just thinking out loud, would anyone actually check for the health of a particular pod?

xinWeiWei24: Good question — generally we don’t focus on the health of a specific pod unless it represents a critical or singleton component, there are edge cases—e.g., pods deployed via custom binaries or standalone images—where they’re not tightly managed by controllers, and in those cases checking individual pod health may still be necessary.

kcl/lib/steps/k8s/validate_workload.k
xinWeiWei24: Yes, this step is meant to be usable in two scenarios: (1) right after a Job is created, to confirm it has come up and is making progress (running + no failed pods); (2) after waiting for the Job to finish, to confirm everything succeeded. As long as the Job is either succeeded or still running (with no Failed condition)

kcl/lib/steps/k8s/validate_workload.k
xinWeiWei24: Done — refactored as suggested. Each case branch now calls check_pods / check_controller_pods directly, and $pods no longer crosses branches.

kcl/lib/steps/k8s/validate_workload.k
xinWeiWei24: Done — added an existence probe (kubectl get ... -o name) at the top of check_controllers and check_jobs. If the resource can't be read (missing / API error / permission denied), it now fails loudly and skips that item instead of silently defaulting fields to 0.

wonderyl: I don't think the existence of the workload directly translate to these fields exist. I was more thinking about when field like completion doesn't exist, just print out an error and exit. The result is not usable when you use default anyway.

kcl/lib/steps/k8s/validate_workload.k
wonderyl: Then what happens when you want to check in scenario 2, it actually meets scenario 1 criteria? It's cleaner just split them into two functions and make the name of the function clear, if you check succeed and active, name name it checkJobSucceedOrActive. Health mean different things in different scenarios.
This also prompt me thinking, do you need a generic check workload health function? since you need to use them separately. one for checking cl2 drivers, which are jobs, one for checking kube-system workloads, which doesn't have job at all.

kcl/lib/steps/k8s/validate_workload.k
wonderyl: If this is only a hypothetical use case, I suggest to remove it for now and add it back when someone really have a use case for it.

kcl/lib/steps/k8s/validate_workload.k
xinWeiWei24: These fields are marked as omitempty in the JSON, so it’s not straightforward to trigger a clear (loud) failure. Also, the job validation logic has been moved to another file and will be included in the next PR.

kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -64,54 +64,50 @@ check_controllers() {
...
-}
-
 # check_pods POD_LIST
-# Scan each pod in a whitespace-separated list. Fails if the list is empty.
+# Check each pod in a whitespace-separated list: status is Running (Ready
+# condition True, all containers ready) or Succeeded, then scan its logs for
+# errorPatterns. Fails if the list is empty.
 check_pods() {
-  local pods=$1 pod
+  local pods=$1 pod phase ready_cond not_ready log_output matches
```
wonderyl: you don't need to define variables before their usage.

kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -0,0 +1,152 @@
...
+      fail "$kind ${namespace}/$n not fully ready ($ready/$desired)"
+    fi
+  done
+}
+
+# check_pods POD_LIST
+# Check each pod in a whitespace-separated list: status is Running (Ready
+# condition True, all containers ready) or Succeeded, then scan its logs for
+# errorPatterns. Fails if the list is empty.
+check_pods() {
```
wonderyl, line 72: maybe for future prs, but I thought you said you will check restarts and oom kills.

xinWeiWei24: Yes, I will add to next PR

wonderyl (approved):

PR #1236, opened 2026-06-25
Record batch machine command timings

karenychen:
## Summary
- record per-batch BatchPutMachine command timings in AKSMachineClient
- emit batch_command_execution_times as scale_machine operation metadata
- cover metric capture and metadata emission in machine-client tests

## Tests
- PYTHONPATH=modules/python python3 -m pytest modules/python/tests/test_aks_machine_client.py modules/python/tests/test_machine_crud.py modules/python/tests/test_crud_main.py
- git diff --check


karenychen:
### Code review

No issues found. Checked for bugs and AGENTS.md compliance.

xuexu6666:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

karenychen:
successful test run: https://dev.azure.com/akstelescope/telescope/_build/results?buildId=72044&view=results

modules/python/clients/aks_machine_client.py
```diff
@@ -905,15 +898,26 @@ def _create_batch_machines(
...
             first_machine_name=first_machine_name,
         )
+        end_time = datetime.now(timezone.utc)
+        execution_time_seconds = (end_time - start_time).total_seconds()
+        request.batch_command_execution_times[first_machine_name] = {
+            "start_time": start_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
+            "end_time": end_time.strftime("%Y-%m-%dT%H:%M:%SZ"),
+            "execution_time_seconds": execution_time_seconds,
+            "total_machines_in_batch": len(chunk),
+        }
```
LeonardCareer, line 915: Should record batch timing even when the BatchPutMachine request fails.

Right now the metric is written only after `_make_batch_request()` returns successfully. If it raises after retries / timeout / non-2xx, `_scale_machine_batch()` catches that worker failure and returns `[]`, so the operation only reports `landed X/Y machines` and the failed batch has no timing entry. That makes the new metric biased toward successful batches and hides the slow/failing batch we most need for debugging.

Could we wrap the request in `try/finally` so every batch records elapsed time, and include a small status/error field?

karenychen: The batch PutMachine command execution time is the cx-observed latency in frontend. If this step fails, we would be more interested in the error than the latency and will be manually looking into internal logs for them.

LeonardCareer: Makes sense, thanks for explaining.

LeonardCareer (approved):
lgtm

PR #1239, opened 2026-07-02
Add k8s job validation steps

xinWeiWei24:
New validate_job.k with two lambdas that emit Azure CLI tasks:
- ValidateJobRunning: checks .status.active >= 1 and Failed condition != True
- ValidateJobCompleted: checks .status.succeeded >= .spec.completions and Complete condition == True

### DIFF
--- a/kcl/lib/steps/k8s/validate_job.k
+++ b/kcl/lib/steps/k8s/validate_job.k
@@ -1,0 +1,102 @@
+import azure_pipelines.ap.steps
+import lib.steps.azure
+
+# ValidateJobRunning validates that a Kubernetes Job is currently running:
+# .status.active >= 1 AND .status.conditions[type=Failed].status != "True".
+# Requires kubectl on the agent and a kubeconfig already in place (e.g. via
+# GetCredentials).
+ValidateJobRunning = lambda \
+    serviceConnection: str, \
+    name: str, \
+    namespace: str = "default", \
+    displayName: str = "", \
+    continueOnError: bool = Undefined \
+    -> steps.Step {
+    assert name != "", "name must be non-empty"
+    title = displayName if displayName else "Validate job/${name} running"
+    script = """
+set +e
+
+failures=0
+fail() { echo "FAILED JOB CHECK: $*"; failures=$((failures + 1)); }
+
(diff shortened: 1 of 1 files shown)


wonderyl (approved):

kcl/lib/steps/k8s/validate_job.k
```diff
@@ -0,0 +1,102 @@
...
+  fail "job ${namespace}/${name} has no active pods"
+fi
+
+if [ "$failures" -gt 0 ]; then
+  echo "Validation failed with $failures error(s) for job ${namespace}/${name}"
+  exit 1
+fi
+echo "job ${namespace}/${name} is running"
+"""
+    azure.AzCli(serviceConnection, title, script, continueOnError=continueOnError)
```
liyu-ma, line 48: I see you only use `kubectl` here, then can we just use `Bash` step for simplicity? And then you don't need to pass in `serviceConnection`, which is only required for AzCli

kcl/lib/steps/k8s/validate_job.k
```diff
@@ -0,0 +1,102 @@
...
+  fail "job ${namespace}/${name} Complete condition='$complete_cond' (expected True)"
+fi
+
+if [ "$failures" -gt 0 ]; then
+  echo "Validation failed with $failures error(s) for job ${namespace}/${name}"
+  exit 1
+fi
+echo "job ${namespace}/${name} completed successfully"
+"""
+    azure.AzCli(serviceConnection, title, script, continueOnError=continueOnError)
```
liyu-ma, line 101: Same as above, I think `steps.Bash` should be sufficient.

kcl/lib/steps/k8s/validate_job.k
xinWeiWei24: oh I think you are right, let me change

xinWeiWei24: oh I think we need to use AzCli since kubectl on AKS needs Azure creds to get a token (kubelogin's azurecli mode reads az account's token), so the step has to run inside an az logined context. AzureCLI@2 handles that via the service connection; a plain Bash step wouldn't be authenticated and kubectl would fail with a credential error.

liyu-ma (approved):

PR #1237, opened 2026-06-25
Add InstallPrometheus and InstallHelm lib steps

LeonardCareer:
## Summary

Adds two new lib steps under `kcl/lib/steps/k8s/`:

- `install_helm.k` — `InstallHelm`: installs Helm v3 on the pipeline agent.
- `install_prometheus.k` — `InstallPrometheus`: installs the [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) Helm chart (Operator + Prometheus + kube-state-metrics + Grafana + Alertmanager) into the current kubectl context.

```kcl
import lib.steps.k8s.install_helm as helm
import lib.steps.k8s.install_prometheus as prom

helm.InstallHelm()
prom.InstallPrometheus(
    serviceConnection = SERVICE_CONNECTION,
    valuesFile        = "kcl/<scenario>/prometheus-values.yaml",
)
```

`InstallHelm` is its own step so other helm-based steps can reuse it.

## Why

Multiple benchmark scenarios need to provision an in-cluster Prometheus on a freshly created AKS cluster — currently each scenario hand-rolls its own helm/kubectl-apply step. This step covers both cases discussed so far:

- **Xinwei**'s 15K-node bench (kube-prometheus-stack with cilium PodMonitor / PrometheusRule and KSM)
- **Leonard**'s S5 lease bench (lightweight: Operator + Prometheus only, all other components disabled in values.yaml)

## What the caller must prepare

1. **Helm on PATH** — call `InstallHelm()` in a prior step.
2. **Working kubeconfig** — call `azure.GetCredentials(...)` in a prior step so kubectl works against the target cluster.
3. **`values.yaml`** checked in to the caller's pipeline repo, e.g. `kcl/<scenario>/prometheus-values.yaml`. The path passed to `valuesFile` is **repo-relative under `$(Pipeline.Workspace)/s/`**. All workload-specific tuning (retention, storage, resource requests, scrape rules, PodMonitor/ServiceMonitor selectors, node selectors, tolerations, enabling/disabling Grafana/Alertmanager/KSM) lives in this file — see the chart's [`values.yaml`](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml) for the full surface.
[...]

PR #1241, opened 2026-07-03
validate_workload: detect OOMKilled and excessive

xinWeiWei24:


kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -3,29 +3,37 @@ import lib.steps.azure
 
 # ValidateWorkloadHealth validates that a Kubernetes controller workload
 # (Deployment, StatefulSet, or DaemonSet) is healthy. It first checks the
-# resource status, then scans the full container logs of its pods for error
-# patterns. Requires kubectl on the agent and a kubeconfig already in place
-# (e.g. via GetCredentials).
+# resource status, then inspects each pod's live status for OOMKilled
```
vittoriasalim, line 6: I am thinking if it is a good idea to do this in prometheus with [kube_pod_container_status_restarts_total metric](https://docs.cloudera.com/management-console/1.5.4/monitoring-metrics/topics/cdppvc_ds_kube_pod_container_status_restarts_total_trics.html) Leonard had a PR for prometheus extension https://github.com/Azure/telescope/pull/1237/changes#diff-62ff5b9ce16ceffb2a778a23f876e84ed90e3970c4d2bedc177fb6bc417e89ed , will this still fit your use case?

xinWeiWei24: Thanks for the suggestion. I’d prefer to keep this test lightweight and avoid adding Prometheus or other extra load. I’m looking for a simpler validation method.

vittoriasalim (approved):

PR #1246, opened 2026-07-07
fix(kperf): support both runkperf and kperf rg run result formats

xinWeiWei24:
- Normalize by unwrapping `.result` when present (runkperf), pass through otherwise (kperf rg run).

### DIFF
--- a/kcl/lib/steps/kperf/collect_result.k
+++ b/kcl/lib/steps/kperf/collect_result.k
@@ -27,18 +27,36 @@ cloud_info=$(jq -n \\
     cluster_id: $cluster_id
   }')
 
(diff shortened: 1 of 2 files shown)


kcl/lib/steps/kperf/collect_result.k
```diff
@@ -27,18 +27,36 @@ cloud_info=$(jq -n \\
     cluster_id: $cluster_id
   }')
 
+# runkperf bench output: { description, loadSpec, result: {...}, info }  → extract .result
```
wonderyl, line 30: double checking, do you need to update yaml files?

wonderyl (approved):

LeonardCareer (approved):
LGTM

PR #1252, opened 2026-07-10
validate_workload: don't fail on scaled-to-zero workloads or vanished pods

xinWeiWei24:


vittoriasalim (approved):

kcl/lib/steps/k8s/validate_workload.k
```diff
@@ -82,6 +82,13 @@ check_pods() {
   local pod
   for pod in $pods; do
     echo "::: checking pod ${namespace}/$pod"
+    # The pod list is a snapshot; a pod may be deleted before we inspect it
+    # (e.g. rollout, eviction, completed Job pod). A gone pod is not a failure,
+    # so skip it instead of reporting an empty phase.
+    if ! kubectl -n "${namespace}" get pod "$pod" -o name >/dev/null 2>&1; then
```
liyu-ma: Would this run into an infinite loop if pods are deleted during the check

wonderyl: It looks like you only use `kubectl get pod`, it is possible to do this? dump all the required info of all pods into a json file, then look up the json file to decide. In this case you only talk to kube-apiserver once, thus avoids the inconsistency.

xinWeiWei24: done

xinWeiWei24: No, but now it changed to checking a file and also won't run into an infinite loop

liyu-ma (approved):

wonderyl (approved):

PR #1272, opened 2026-07-29
ci: trigger tf integration via /run-tf-integration comment

LeonardCareer:
## What

Change the `Terraform Integration` workflow so it only runs when a telescope reviewer explicitly triggers it, instead of running on every push.

## Why

Every fork PR push currently fans out into ~62 privileged plan jobs, each of which requests a `terraform-privileged` environment approval → dozens of emails to reviewers per push. Team consensus (scrum): move the trigger under reviewer control.

## How

- **`terraform-integration.yml`**
  - Trigger switched from `pull_request_target` to `issue_comment: [created]`
  - Every privileged job gates on `github.event.issue.pull_request != null && github.event.comment.body == '/run-tf-integration'`
  - Checkout uses `refs/pull/{n}/head` (no PR head sha in `issue_comment` payload)
  - `setup-matrix` explicitly `git fetch origin main` before diffing (base branch is not auto-fetched under `issue_comment`)
  - New `report-status` job posts a commit status back to the PR head sha so the run shows up in the PR checks section

- **`tf-plan-comment.yml`** (new)
  - On PR open / synchronize / reopen, posts (or refreshes) a sticky comment telling telescope reviewers how to trigger.

## Security

- `issue_comment` workflows load from the base branch (main), so a fork PR cannot inject workflow code.
- The `terraform-privileged` environment gate remains, so even if a fork PR author comments the slash command, secrets are not released until an environment reviewer approves. Same security model as post-MSRC #127928.

## Test plan

- [ ] PR opened → sticky comment appears
- [ ] Reviewer comments `/run-tf-integration` on tf PR → matrix jobs run after env approval
- [ ] Reviewer comments `/run-tf-integration` on non-tf PR → setup-matrix returns null, plan skipped, report-status = success
- [ ] Failing scenario → report-status = failure, shows in PR checks

Related: MSRC #127928.


.github/workflows/terraform-integration.yml
```diff
@@ -19,7 +15,32 @@ env:
   TERRAFORM_AWS_MODULES_DIR: modules/terraform/aws
 
 jobs:
+  gate:
+    if: github.event.review.state == 'approved'
```
wonderyl: @liyu-ma, here is the dilemma. Suppose I'm a reviewer, I can't approve before I feel confident the change is correct, then I need to check the tf change is working. If I already checked it, the test is not useful anymore. 
Alternatively. If this remains a pr check, a reviewer can trigger it before he starts reviewing and check back the result later.

.github/workflows/add-tf-integration-comment.yml
```diff
@@ -0,0 +1,36 @@
+name: Terraform Plan Trigger Comment
```
wonderyl: nit: should it be called "tf-integration-comment" since it triggers the tf-intgration workflow

.github/workflows/terraform-integration.yml
```diff
@@ -175,3 +174,45 @@ jobs:
         env:
           ARM_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
           CLOUD: ${{ matrix.cloud }}
+
+  report-status:
+    needs: [terraform-test, setup-matrix, terraform-plan]
+    if: always()
```
liyu-ma: Do we want `always()`, or use the same condition as other jobs
```
github.event.issue.pull_request != null &&
github.event.comment.body == '/run-tf-integration'
```

.github/workflows/add-tf-integration-comment.yml
wonderyl: One more, consider naming it as an action, since it's an workflow. e.g. adds-tf-integration-comment.yml

.github/workflows/terraform-integration.yml
```diff
@@ -20,6 +13,9 @@ env:
 
 jobs:
   terraform-test:
+    if: |
+      github.event.issue.pull_request != null &&
```
wonderyl, line 17: If I understand correctly the reviewer now need to do 2 things, 1. adds the /run-tf-integration comment and also authorize the workflows to use terraform-privileged env. 
Is it possible to reduce it to one? 
E.g. check who make the /run-tf-integration comment and only trigger the workflows when they are from a allow list. Then it triggers workflow runs without privileged env, since it's triggered from authorized personals.

.github/workflows/terraform-integration.yml
wonderyl: it depends on terraform-test and terraform-plan, it can't start until they finished.

.github/workflows/terraform-integration.yml
LeonardCareer: Yeah, it makes sense. 
Just for one step "check who made the comment", it needs to be hardcoded as `contains(fromJSON('["wonderyl","liyu-ma","xinWeiWei24","vittoriasalim","LeonardCareer"]'), github.event.comment.user.login)`, since option 1) the default GITHUB_TOKEN doesn't have admin:org scope to query team membership. option 2) Using a PAT would require rotation every ~7 days per Azure org policy, and option 3) a GitHub App needs org admin approval to install.

.github/workflows/add-tf-integration-comment.yml
LeonardCareer: Yeah, will do. thanks for the suggestions.

.github/workflows/terraform-integration.yml
LeonardCareer: As suggested, using main branch codeowner to verify whether the commenter is authorized.

.github/workflows/terraform-integration.yml
```diff
@@ -176,8 +193,8 @@ jobs:
           CLOUD: ${{ matrix.cloud }}
 
   report-status:
-    needs: [terraform-test, setup-matrix, terraform-plan]
-    if: always()
+    needs: [gate, terraform-test, setup-matrix, terraform-plan]
+    if: always() && needs.gate.result == 'success'
```
wonderyl, line 194: is `always() && needs.gate.result == 'success'` equivalent to `needs.gate.result == 'success'`

.github/workflows/terraform-integration.yml
```diff
@@ -60,17 +75,20 @@ jobs:
...
       - name: Checkout PR head
         uses: actions/checkout@v4
         with:
-          repository: ${{ github.event.pull_request.head.repo.full_name }}
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: refs/pull/${{ github.event.issue.number }}/head
           fetch-depth: 0
           persist-credentials: false
 
+      - name: Fetch main branch for diff
```
wonderyl: what is this used for?

.github/workflows/terraform-integration.yml
```diff
@@ -175,3 +191,45 @@ jobs:
         env:
           ARM_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
           CLOUD: ${{ matrix.cloud }}
+
+  report-status:
+    needs: [gate, terraform-test, setup-matrix, terraform-plan]
```
wonderyl, line 193: nit: IIUC the dependency looks like the one below. So report-status only need to depend on terraform-test and terraform-plan
```
gate <- terraform-test <------------------- report-status
   ^--- setup matrix <--- terraform-plan <--------|
```

.github/workflows/terraform-integration.yml
LeonardCareer: Good catch. When switching the trigger from pull_request_target to issue_comment, I was worried that actions/checkout might not fetch the base branch anymore. So I added this explicit fetch defensively.

Just verified actions/checkout@v4 with fetch-depth: 0 already runs git fetch --prune origin +refs/heads/*:refs/remotes/origin/*, which pulls all branches including main. So origin/main is available automatically and this extra fetch is redundant.

Removing it.

.github/workflows/terraform-integration.yml
LeonardCareer: Not equivalent to always() alone either. Just verified:
If only needs.gate.result == 'success', GitHub's default "skip when needs fails" fires before the if is evaluated, so when terraform-plan actually fails, report-status gets silently skipped and no failure status is posted. 

Similar question, couldwe only use always here? No, for any other comment beside "/run-tf-integration", all job is skipped, and **report-status** will post a fake succussed.

.github/workflows/terraform-integration.yml
LeonardCareer: Yeah, I understand their relationship. But we use needs.setup-matrix.result and needs.gate.result below. I checked — when this job runs, gate and setup-matrix are actually finished (as ancestors of terraform-test/terraform-plan), and it works at runtime. However, actionlint flags it as a grammar error since they're not in needs. I prefer to keep them explicit.

wonderyl (approved):

PR #1281, opened 2026-08-06
feat(kcl): add AKS cloud-node-manager validation step

xinWeiWei24:
Add a reusable step that verifies the cloud-node-manager component: checks all pods are Running/Ready and that node providerIDs are set, confirming the cloud provider integration is healthy.


kcl/lib/steps/azure/validate_aks_cloud_node_manager.k
```diff
@@ -0,0 +1,49 @@
...
+  bad=$(printf '%s\\n' "$pods" | awk '{split($2,r,"/"); if ($3 != "Running" || r[1] != r[2]) print $1}')
+  if [ -n "$bad" ]; then
+    fail "cloud-node-manager pods not Running/Ready: $bad"
+  fi
+fi
+
+# Check every node in a single list call so no node is missed; a per-node
+# "kubectl get node" would be one call per node and is too slow on large
+# clusters. All nodes are scanned locally; only problem nodes are printed.
+node_table=$(kubectl get nodes --no-headers -o custom-columns='NAME:.metadata.name,READY:.status.conditions[?(@.type=="Ready")].status,PROVIDER:.spec.providerID' 2>/dev/null)
```
wonderyl: describe ds has some status to use 
```
k describe ds/cloud-node-manager -n kube-system
Name:           cloud-node-manager
Namespace:      kube-system
Selector:       k8s-app=cloud-node-manager
Node-Selector:  <none>
Labels:         addonmanager.kubernetes.io/mode=Reconcile
                k8s-app=cloud-node-manager
                kubernetes.azure.com/managedby=aks
                kubernetes.io/cluster-service=true
Annotations:    deprecated.daemonset.template.generation: 1
Desired Number of Nodes Scheduled: 7
Current Number of Nodes Scheduled: 7
Number of Nodes Scheduled with Up-to-date Pods: 7
Number of Nodes Scheduled with Available Pods: 7
Number of Nodes Misscheduled: 0
```

kcl/lib/steps/azure/validate_aks_cloud_node_manager.k
```diff
@@ -0,0 +1,49 @@
...
+    script = """
+set +e
+failures=0
+fail() { echo "FAILED CLOUD-NODE-MANAGER CHECK: $*"; failures=$((failures + 1)); }
+
+pods=$(kubectl -n kube-system get pods -l k8s-app=cloud-node-manager --no-headers 2>/dev/null)
+if [ -z "$pods" ]; then
+  fail "no cloud-node-manager pods found"
+else
+  bad=$(printf '%s\\n' "$pods" | awk '{split($2,r,"/"); if ($3 != "Running" || r[1] != r[2]) print $1}')
```
wonderyl: does using `printf "$pods"` work?

kcl/lib/steps/azure/validate_aks_cloud_node_manager.k
```diff
@@ -0,0 +1,49 @@
+import azure_pipelines.ap.steps
+
+# cloud-node-manager: every node is Ready and carries a providerID (set by the
+# cloud provider integration).
+ValidateCloudNodeManager = lambda serviceConnection: str, displayName: str = "", continueOnError: bool = Undefined -> steps.Step {
```
wonderyl, line 7: Is this specific to CNM or you can generalize it to all DaemonSets?

kcl/lib/steps/azure/validate_aks_cloud_node_manager.k
```diff
@@ -0,0 +1,49 @@
...
+    title = displayName if displayName else "Validate cloud-node-manager"
+    script = """
+set +e
+failures=0
+fail() { echo "FAILED CLOUD-NODE-MANAGER CHECK: $*"; failures=$((failures + 1)); }
+
+pods=$(kubectl -n kube-system get pods -l k8s-app=cloud-node-manager --no-headers 2>/dev/null)
```
wonderyl: It seems to me the `k describe ds` has ready pods already, do may not need to use get pods.

kcl/lib/steps/azure/validate_aks_cloud_node_manager.k
xinWeiWei24: This check is specific to CNM. The providerID will be set only when CNM is healthy. The other components will have their own validation checks in following PRs.

kcl/lib/steps/azure/validate_aks_cloud_node_manager.k
xinWeiWei24: Agreed — replaced the get pods parsing with the DaemonSet status so we no longer list pods.

kcl/lib/steps/azure/validate_aks_cloud_node_manager.k
xinWeiWei24: Good idea — switched to the DaemonSet status (ready/available/updated/misscheduled vs desired) and dropped the pod parsing.

kcl/lib/steps/azure/validate_aks_cloud_node_manager.k
wonderyl: Then does it make sense to split this into two functions, one is check a generic ds, it's reusable, the other only checks provider id.

kcl/lib/steps/azure/validate_aks_cloud_node_manager.k
```diff
@@ -1,43 +1,52 @@
...
+# cloud-node-manager's job is to stamp each Node with a providerID; verify that
+# in a single list call so no node is missed and large clusters stay fast.
+node_table=$(kubectl get nodes --no-headers -o custom-columns='NAME:.metadata.name,PROVIDER:.spec.providerID' 2>/dev/null)
 if [ -z "$node_table" ]; then
   fail "no nodes found"
-fi
-
-report=$(printf '%s\\n' "$node_table" | awk '
+else
+  report=$(printf '%s\\n' "$node_table" | awk '
```
wonderyl: what's the function of `\n`?

xinWeiWei24: Command substitution $(...) strips trailing newlines, so $node_table has no newline after the last node. printf '%s\n' re-adds one so awk sees the final line as a complete record (and the piped output is clean). printf '%s' alone would merge/drop the last line.

wonderyl: Is this true? when I verified it. It doesn't make a difference
```bash
❯ node_table=$(kubectl get nodes --no-headers -o custom-columns='NAME:.metadata.name,PROVIDER:.spec.providerID' 2>/dev/null)

~ via  v22.22.2 took 2s
❯ printf '%s\\n' "$node_table" | awk '
{
  total++
  if ($2 == "<none>" || $2 == "") { print "node " $1 " missing providerID"; bad++ }
}
END { print "checked " total " nodes, " bad+0 " problem(s)"; exit (bad > 0) }'
checked 7 nodes, 0 problem(s)

~ via  v22.22.2
❯ printf "$node_table" | awk '
{
  total++
  if ($2 == "<none>" || $2 == "") { print "node " $1 " missing providerID"; bad++ }
}
END { print "checked " total " nodes, " bad+0 " problem(s)"; exit (bad > 0) }'
checked 7 nodes, 0 problem(s)

```

xinWeiWei24: Thanks for pointing this out!!!! i change to directly echo

wonderyl (approved):

PR #1282, opened 2026-08-07
feat(validate): add metrics-server AKS component validator

xinWeiWei24:
Verifies metrics-server by sampling `kubectl top node` (first 50 nodes) and asserting each returns a CPU reading. Controller/pod health is left to ValidateWorkloadHealth at the pipeline level.


liyu-ma (approved):
LGTM. Do you think we should also check metrics-server pod ready at the same time, in case they are flapping?

wonderyl (approved):

PR #1284, opened 2026-08-11
Use 16-core VM for cl2pool in hyperscale pod scheduling pipeline

vittoriasalim:
Updates the cl2pool node SKU from Standard_D8S_v4 to Standard_D16S_v4 (16 cores) in the hyperscale pod scheduling pipeline, targeting the v2 branch. Regenerated pipeline.yaml with `kcl run`.


xinWeiWei24 (approved):
LGTM

PR #1286, opened 2026-08-13
feat: add Azure CNS validation step

xinWeiWei24:
Adds an Azure CNS validation step to verify pod IP allocation and in-cluster service connectivity.

### DIFF
--- a/kcl/lib/steps/azure/validate_aks_cns.k
+++ b/kcl/lib/steps/azure/validate_aks_cns.k
@@ -10,9 +10,10 @@ set +e
 
 pod="aks-validation-cns-$RANDOM"
 trap 'kubectl delete pod "$pod" --ignore-not-found >/dev/null 2>&1' EXIT
-# Verify the pod's dataplane can reach an in-cluster service (the kubernetes
-# API ClusterIP), proving CNI IP allocation and routing work end-to-end. Use a
-# plain connect because busybox nc has no reliable -z.
+
+# Verify CNI IP allocation and routing to the kubernetes API ClusterIP.
+# Kubernetes provides the API service address through KUBERNETES_SERVICE_HOST
+# and KUBERNETES_SERVICE_PORT. Redirect stdin with /dev/null so nc exits after connecting.
 reach_out=$(kubectl run "$pod" --image=mcr.microsoft.com/azurelinux/busybox:1.36 --restart=Never -i --rm --pod-running-timeout=120s \
   --command -- sh -c 'nc -w5 "$KUBERNETES_SERVICE_HOST" "$KUBERNETES_SERVICE_PORT" </dev/null && echo REACHABLE' 2>&1)
 echo "dataplane check: $reach_out"
--- a/kcl/lib/steps/azure/validate_aks_cns.k
+++ b/kcl/lib/steps/azure/validate_aks_cns.k
@@ -1,0 +1,43 @@
+import azure_pipelines.ap.steps
+
+# azure-cns: a new pod gets an IP from the CNI IPAM, and the pod dataplane can
+# reach in-cluster services. For daemonset/pod health, pair with
+# ValidateWorkloadHealth(kind="daemonset", name="azure-cns", namespace="kube-system").
+ValidateAzureCNS = lambda serviceConnection: str, displayName: str = "", continueOnError: bool = Undefined -> steps.Step {
+    title = displayName if displayName else "Validate azure-cns"
+    script = """
+set +e
+failures=0
+fail() { echo "FAILED AZURE-CNS CHECK: $*"; failures=$((failures + 1)); }
+
+pod="aks-validation-cns-$RANDOM"
+trap 'kubectl delete pod "$pod" --ignore-not-found >/dev/null 2>&1' EXIT
+kubectl run "$pod" --image=mcr.microsoft.com/azurelinux/busybox:1.36 --restart=Never --command -- sleep 300
+kubectl wait --for=condition=Ready pod/"$pod" --timeout=120s
+if [ $? -ne 0 ]; then
+  fail "test pod did not become Ready (IP allocation may have failed)"
+  echo "azure-cns validation failed with $failures error(s)"; exit 1
+fi
+
+pod_ip=$(kubectl get pod "$pod" -o jsonpath='{.status.podIP}')
+echo "pod IP: $pod_ip"
+if [ -z "$pod_ip" ]; then
+  fail "pod did not receive an IP from azure-cns"
+fi
+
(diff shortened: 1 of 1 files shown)


liyu-ma (approved):

kcl/lib/steps/azure/validate_aks_cns.k
```diff
@@ -0,0 +1,43 @@
...
+ValidateAzureCNS = lambda serviceConnection: str, displayName: str = "", continueOnError: bool = Undefined -> steps.Step {
+    title = displayName if displayName else "Validate azure-cns"
+    script = """
+set +e
+failures=0
+fail() { echo "FAILED AZURE-CNS CHECK: $*"; failures=$((failures + 1)); }
+
+pod="aks-validation-cns-$RANDOM"
+trap 'kubectl delete pod "$pod" --ignore-not-found >/dev/null 2>&1' EXIT
+kubectl run "$pod" --image=mcr.microsoft.com/azurelinux/busybox:1.36 --restart=Never --command -- sleep 300
```
wonderyl: Maybe not in thsi pr, but I'm thinking how we can update busybox version in the future. Would it be useful to use a constant?

kcl/lib/steps/azure/validate_aks_cns.k
```diff
@@ -0,0 +1,43 @@
...
+pod_ip=$(kubectl get pod "$pod" -o jsonpath='{.status.podIP}')
+echo "pod IP: $pod_ip"
+if [ -z "$pod_ip" ]; then
+  fail "pod did not receive an IP from azure-cns"
+fi
+
+# Verify the pod's dataplane can reach an in-cluster service (the kubernetes
+# API ClusterIP), proving CNI routing works end-to-end, not just IPAM. Use a
+# plain connect (busybox nc has no reliable -z) and gate on its exit code.
+reach_out=$(kubectl exec "$pod" -- sh -c 'nc -w5 "$KUBERNETES_SERVICE_HOST" "$KUBERNETES_SERVICE_PORT" </dev/null && echo REACHABLE' 2>&1)
```
wonderyl: I wonder why do run the command separately, instead of passing it with `--command`?

kcl/lib/steps/azure/validate_aks_cns.k
wonderyl: In last time PR, you used `--pod-running-timeout`, but not this one, why?

xinWeiWei24: Good point. I’ll move the shared image version to a constant in a follow-up PR.

xinWeiWei24: --pod-running-timeout is only useful when kubectl run waits for and attaches to the pod, such as with -i --rm in the CoreDNS check. Without that attached flow, kubectl run returns immediately after creating the pod. Here, we wait for readiness separately using kubectl wait --timeout=120s, so --pod-running-timeout is not needed. I’ve updated the condition to make this clearer.

kcl/lib/steps/azure/validate_aks_cns.k
xinWeiWei24: We keep the pod running because this check validates two separate things: first that CNS allocates a pod IP and the pod becomes Ready, and then that its dataplane can reach the Kubernetes service. Passing nc directly with --command would make it a one-off pod and combine both checks, so we would lose the explicit IP allocation check

wonderyl: If the pod has no ip, would it be able to run nc?

xinWeiWei24: You’re right. Without an IP, the pod would not start and nc would not run, so running nc directly would implicitly validate IP allocation as well.

kcl/lib/steps/azure/validate_aks_cns.k
```diff
@@ -0,0 +1,25 @@
...
+    script = """
+set +e
+
+pod="aks-validation-cns-$RANDOM"
+trap 'kubectl delete pod "$pod" --ignore-not-found >/dev/null 2>&1' EXIT
+# Verify the pod's dataplane can reach an in-cluster service (the kubernetes
+# API ClusterIP), proving CNI IP allocation and routing work end-to-end. Use a
+# plain connect because busybox nc has no reliable -z.
+reach_out=$(kubectl run "$pod" --image=mcr.microsoft.com/azurelinux/busybox:1.36 --restart=Never -i --rm --pod-running-timeout=120s \
+  --command -- sh -c 'nc -w5 "$KUBERNETES_SERVICE_HOST" "$KUBERNETES_SERVICE_PORT" </dev/null && echo REACHABLE' 2>&1)
```
wonderyl: where is $KUBERNETES_SERVICE_HOST and $KUBERNETES_SERVICE_PORT defined?

wonderyl: what is `</dev/null` used for?

xinWeiWei24: They are automatically injected into every pod by Kubernetes through service environment variables for the default/kubernetes Service. They point to the Kubernetes API Service ClusterIP and port.

xinWeiWei24: Since kubectl run -i keeps stdin open, nc could wait for input after connecting. Redirecting stdin from [null](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) gives it an immediate EOF, so it only tests whether the TCP connection can be established and then exits.

wonderyl:
```
❯ time nc -w10 www.bing.com 80 </dev/null && echo REACHABLE

real    0m11.818s
user    0m0.000s
sys     0m0.004s
REACHABLE


❯ time nc -w5 www.bing.com 80 </dev/null && echo REACHABLE

real    0m5.021s
user    0m0.002s
sys     0m0.000s
REACHABLE

❯ time nc -w5 www.bing.com 80 && echo REACHABLE

real    0m6.863s
user    0m0.002s
sys     0m0.000s
REACHABLE
```

It doesn't seem /dev/null has any effect, it will wait for the timeout regardless, and not having it won't stop the command from finishing.

xinWeiWei24: You’re right. I misunderstood the BusyBox nc behavior here. The redirection does not avoid the -w5 wait, so I removed it. Thanks for verifying it.

wonderyl (approved):

PR #1296, opened 2026-08-20
Add AKS konnectivity-agent validation step

xinWeiWei24:
Functional validator that execs into a temporary busybox pod to verify the control-plane-to-node konnectivity tunnel works.

### DIFF
--- a/kcl/lib/steps/azure/validate_aks_konnectivity_agent.k
+++ b/kcl/lib/steps/azure/validate_aks_konnectivity_agent.k
@@ -1,0 +1,29 @@
+import azure_pipelines.ap.steps
+import lib.const
+
+# konnectivity-agent: the tunnel from control plane to nodes works, exercised
+# via kubectl exec against a temporary pod. For pod/controller health, pair with
+# ValidateWorkloadHealth(kind="deployment", name="konnectivity-agent", namespace="kube-system").
+ValidateKonnectivityAgent = lambda serviceConnection: str, displayName: str = "", continueOnError: bool = Undefined -> steps.Step {
+    title = displayName if displayName else "Validate konnectivity-agent"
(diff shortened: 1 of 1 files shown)


liyu-ma (approved):

kcl/lib/steps/azure/validate_aks_konnectivity_agent.k
```diff
@@ -0,0 +1,29 @@
+import azure_pipelines.ap.steps
```
wonderyl, line 1: I think this file should live in lib/steps/k8s, since it's more related to k8s. Though it uses AzCli, this can easily adapt to other cloud.

I don't think I noticed this before, so it may apply to your past PRs.

xinWeiWei24: Thanks, I was thinking the same thing. A few of these checks are Azure-specific (azure-cns, and later the CSI checks), while others are pure Kubernetes (coredns, kube-proxy). But I agree the underlying logic is Kubernetes-centric and portable.

Happy to move them. Should I move everything to lib/steps/k8s, or leave the Azure-specific checks (for example, azure-cns) here?

wonderyl: Another option is to group your health check related lambdas into a new folder, since they are likely to be used together. It can be done in another pr, I'll approve this.

wonderyl (approved):

kcl/lib/steps/azure/validate_aks_konnectivity_agent.k
xinWeiWei24: Thanks! will move to new folder in next PR

PR #1297, opened 2026-08-21
refactor(kcl): group health-check validators into lib/steps/health

xinWeiWei24:
Move the 7 component checks + validate_workload/validate_job into a new health/ package and drop the redundant aks_ filename prefix.


liyu-ma (approved):
`validation` may sound a better package name than `health`? Just for your thought.
Image

wonderyl (approved):

PR #1302, opened 2026-08-25
Add ClusterIP service routing health check

xinWeiWei24:
Adds ValidateClusterIPService: deploys a backend behind a ClusterIP and verifies it's reachable from a disposable client pod. Works with any service proxy (kube-proxy, Cilium eBPF); auto-cleans up all resources.

### DIFF
--- a/kcl/lib/steps/health/validate_kube_proxy.k
+++ b/kcl/lib/steps/health/validate_kube_proxy.k
@@ -1,0 +1,77 @@
+import azure_pipelines.ap.steps
+import lib.const
+import lib.steps.azure
+
+# kube-proxy: a ClusterIP service is reachable from within the cluster,
+# exercising service routing. For daemonset/pod health, pair with
+# ValidateWorkloadHealth(kind="daemonset", name="kube-proxy", namespace="kube-system")
+# only on clusters that run kube-proxy; skip it when a replacement (e.g. Cilium
+# eBPF) is used, since there is no kube-proxy daemonset to check.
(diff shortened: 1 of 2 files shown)


liyu-ma (approved):

kcl/lib/steps/health/validate_kube_proxy.k
```diff
@@ -0,0 +1,77 @@
+import azure_pipelines.ap.steps
+import lib.const
+import lib.steps.azure
+
+# kube-proxy: a ClusterIP service is reachable from within the cluster,
+# exercising service routing. For daemonset/pod health, pair with
+# ValidateWorkloadHealth(kind="daemonset", name="kube-proxy", namespace="kube-system")
+# only on clusters that run kube-proxy; skip it when a replacement (e.g. Cilium
```
wonderyl: does it rely on the caller to decide whether the cluster is using kube-proxy?

wonderyl: looks like the logic does not bind to kube proxy, anything that backs a cluster ip can be validated by this, right?

xinWeiWei24: Good catch. The check isn't kube-proxy specific — it validates ClusterIP service routing, which works regardless of the proxy implementation (kube-proxy, Cilium eBPF, etc.). Renamed it to ValidateClusterIPService and updated the comment

wonderyl (approved):

PR #1304, opened 2026-08-29
fix(instrumentation): exclude admission/queue-wait from provisioning metrics (AKS + Machine API)

xuexu6666:
## Problem

Telescope over-reports provisioning latency because it bills time spent waiting to be **admitted** by ARM against the operation itself. This is **not GPU-specific** — it affects multiple pipelines:

- **AKS node-pool CRUD** (create / scale / progressive; GPU **and** non-GPU) via `begin_create_or_update_with_retry`: ARM returns `409 OperationNotAllowed` while a *previous* operation on the cluster is still running, and the sleep-retry loop runs **inside** the timed region. Real example (H100 scale_up, run `78200-2d6bbfea`): reported `606s`, ~180s of which was pre-accept queue-wait.
- **Machine API scale** (`aks_machine_client`) via `BatchPutMachine`: the `429` throttle backoff runs inside `command_execution_time`.
- **EKS**: unaffected (boto3 waiters, no pre-accept serialization).

### AKS node pool — measure from the accepted (2xx) attempt; exclude the queue-wait from `duration`
- `begin_create_or_update_with_retry` stamps each attempt's start and returns `(request_started_at, retry_occurred)` for the attempt ARM **accepts**. A `409 OperationNotAllowed` raises *before* a poller exists, so its stamp is discarded.
- `instrument_nodepool_provisioning` measures `command_execution_time` / `node_readiness_time` from `request_started_at` (counts the accepted request's own frontend time + async provisioning; the failed prior attempts are excluded as queue-wait).
- The queue-wait is excluded from the operation **`duration`** via `Operation.exclude_time()` and recorded (clamped) as `in_progress_wait_seconds`. **Failure paths carry the timing too:** accepted-then-failed and K8s-readiness-failed exclude the pre-accept wait before re-raising; an exhausted never-admitted `OperationNotAllowed`/`EtagMismatch` anchors the exclusion at give-up; a **non-retryable** terminal error (400/403/404) anchors nothing (real ARM latency retained).

### Machine API — exclude the throttle from `command_execution_time` only (NOT `duration`)
- Each batch chunk's total `429` backoff is recorded with its finish time as `(finish, backoff)` in `request.chunk_throttle` — on **success and failure** (recorded in a `finally`, so a chunk that exhausts its retries still contributes).
- `utils.throttle_makespan_delay` computes the throttle's delay to the concurrent PUT-phase makespan: `makespan − max(finishᵢ − backoffᵢ)`. Off-critical-path backoff (a chunk that slept but wasn't last to finish) contributes 0.
- `scale_machine` subtracts that **only from `command_execution_time`** (`max(0, command_end − command_t0 − throttle_wait)`) and records `throttle_wait_seconds`. It does **not** call `Operation.exclude_time`, so the operation `duration` is unchanged — the PUT-phase critical path can differ from the end-to-end one (the last-Ready machine may be a different chunk), so PUT-phase throttle doesn't necessarily delay completion. *(Per @karenychen's review.)*

> Note the intentional asymmetry: the single-PUT AKS path excludes queue-wait from `duration` (the pre-accept wait strictly delays everything downstream — one critical path); the concurrent multi-PUT Machine path scopes the exclusion to `command_execution_time`.

## Files
- `crud/operation.py` — `exclude_time()` primitive + clamped duration adjustment (used by the AKS path; `Operation.end()` is the single owner of `in_progress_wait_seconds`).
- `utils/provisioning_instrumentation.py` — accept-based AKS timing + failure-path timing anchors.
- `utils/common.py` — `throttle_makespan_delay()` critical-path helper.
- `clients/aks_machine_client.py` — per-chunk `(finish, backoff)` recording + scoped `command_execution_time` throttle exclusion.

## Tests
`throttle_makespan_delay` cases (off-critical-path → 0 / tie / runner-up cap); accept-based timing incl. failure-path exclusion (K8s-fail, accepted-then-fail, never-admitted anchor, non-retryable → no anchor); per-chunk backoff-on-failure recording; `Operation` excess-wait clamp. `pytest` green (pre-existing `test_log_gpu_mode_console_echo` is Python-3.10-only `assertNoLogs`, unrelated); pylint clean on changed files.

### DIFF
--- a/modules/python/crud/operation.py
+++ b/modules/python/crud/operation.py
@@ -9,7 +9,7 @@ metadata for operations performed in the Telescope project.
 import json
 import os
 import traceback
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from typing import Dict, Any, Optional
 
 from utils.logger_config import get_logger, setup_logging
@@ -39,6 +39,7 @@ class Operation:
         self.start_timestamp = None
         self.end_timestamp = None
         self.duration = None
+        self.excluded_seconds = 0.0
         self.unit = "seconds"
         self.success = True
         self.error_message = None
@@ -50,6 +51,12 @@ class Operation:
         Start the operation by recording the current time.
         """
         self.start_timestamp = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+
+    def exclude_time(self, seconds: float) -> None:
+        """Accumulate wall-clock ``seconds`` to drop from the duration (e.g. ARM
+        admission/queue-wait). Positive values accumulate; others are ignored."""
+        if seconds is not None and seconds > 0:
+            self.excluded_seconds += seconds
 
     def end(self, success: bool = True, error: Optional[Exception] = None) -> None:
         """
@@ -70,6 +77,13 @@ class Operation:
                 end_dt = datetime.fromisoformat(
                     self.end_timestamp.replace("Z", "+00:00")
                 )
+                if self.excluded_seconds > 0:
+                    excluded = min(
+                        self.excluded_seconds, (end_dt - start_dt).total_seconds()
+                    )
+                    start_dt = start_dt + timedelta(seconds=excluded)
+                    self.start_timestamp = start_dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+                    self.metadata["in_progress_wait_seconds"] = excluded
                 self.duration = (end_dt - start_dt).total_seconds()
             except Exception:
                 # If there's any issue parsing timestamps, don't set duration
--- a/modules/python/utils/common.py
+++ b/modules/python/utils/common.py
@@ -2,6 +2,7 @@ import re
 import os
 import json
 import argparse
+from typing import List, Tuple
 from utils.logger_config import get_logger, setup_logging
 
 # Configure logging
@@ -71,3 +72,20 @@ def str2bool(val):
     if val.lower() in ("false", "no", "0"):
         return False
     raise argparse.ArgumentTypeError("Boolean value expected.")
+
+
+def compute_throttle_delay(chunks: List[Tuple[float, float]]) -> float:
+    """Throttle delay to a concurrent command's completion.
+
+    ``chunks`` is per-worker ``(finish_time, backoff_seconds)``. Backoff is pure
+    delay on a worker's own timeline, so its throttle-free finish is
+    ``finish - backoff`` and the delay is ``max(finish) - max(finish - backoff)``.
+    Only backoff on the critical path counts -- off-path backoff hides behind
+    other workers. Assumes workers start concurrently (each Machine chunk gets its
+    own pool slot); queued workers would make this under-count.
+    """
+    if not chunks:
+        return 0.0
+    actual_finish = max(finish for finish, _ in chunks)
+    finish_without_backoff = max(finish - backoff for finish, backoff in chunks)
+    return max(0.0, actual_finish - finish_without_backoff)
--- a/modules/python/utils/provisioning_instrumentation.py
+++ b/modules/python/utils/provisioning_instrumentation.py
@@ -14,6 +14,29 @@ from utils.logger_config import get_logger
 
 logger = get_logger(__name__)
 
+IN_PROGRESS_CODES = ("OperationNotAllowed", "EtagMismatch")
+MAX_RETRIES = 10               # in-progress (409) submit attempts
+RETRY_WAIT_SECONDS = 30        # sleep between rejected submit attempts
+POLL_INTERVAL_SECONDS = 30     # poller.done() polling cadence
+TIMEOUT_SECONDS = 1800         # LRO poll timeout (slow A100 MIG provisioning)
+
+
+def _record_excluded_time(op, command_start, start_time):
+    """Exclude pre-accept queue-wait from the op duration; return the seconds excluded."""
+    if command_start is None:
+        return 0.0
+    excluded = max(0, command_start - start_time)
+    op.exclude_time(excluded)
+    return excluded
+
+
+def _log_provisioning_success(node_pool_name, label, command_time, readiness_time, queue_wait):
+    logger.info(
+        "[%s] %sARM %.2fs, K8s ready %.2fs | Delta %.2fs | queue-wait excluded %.2fs",
+        node_pool_name, label, command_time, readiness_time,
+        abs(command_time - readiness_time), queue_wait,
+    )
+
 
 def instrument_nodepool_provisioning(
     node_pool_name,
@@ -61,24 +84,23 @@ def instrument_nodepool_provisioning(
             "Concurrent operation failed after %.2fs - ARM: %s, K8s readiness: %s",
             elapsed, arm_status, k8s_status
         )
-        if arm_exc:
-            raise arm_exc
-        raise k8s_exc
+        if not arm_exc:
+            (command_start, _), _ = arm_future.result()
+        else:
+            command_start = getattr(arm_exc, "request_started_at", None)
+        _record_excluded_time(op, command_start, start_time)
+        raise arm_exc or k8s_exc
 
-    arm_result, arm_timestamp = arm_future.result()
+    (command_start, retry_occurred), arm_timestamp = arm_future.result()
     ready_nodes, ready_timestamp = k8s_future.result()
+    command_execution_time = max(0, arm_timestamp - command_start)
+    node_readiness_time = max(0, ready_timestamp - command_start)
 
-    node_readiness_time = ready_timestamp - start_time
-    command_execution_time = arm_timestamp - start_time
-
+    queue_wait = _record_excluded_time(op, command_start, start_time)
     op.add_metadata("node_readiness_time", node_readiness_time)
     op.add_metadata("command_execution_time", command_execution_time)
-    op.add_metadata("retry_occurred", bool(arm_result))
-    logger.info(
-        "[%s] %sARM completed in %.2fs, K8s nodes ready in %.2fs | Delta: %.2fs",
-        node_pool_name, label, command_execution_time, node_readiness_time,
-        abs(command_execution_time - node_readiness_time)
-    )
+    op.add_metadata("retry_occurred", retry_occurred)
+    _log_provisioning_success(node_pool_name, label, command_execution_time, node_readiness_time, queue_wait)
 
     return ready_nodes
 
@@ -90,21 +112,21 @@ def begin_create_or_update_with_retry(
     node_pool_name,
     parameters,
     label="",
-    retries=10,
-    retry_wait=30,
-    poll_interval=30,
-    timeout=1800,
 ):
     """
-    Call begin_create_or_update with retry on OperationNotAllowed/EtagMismatch,
-    polling every poll_interval seconds and raising TimeoutError after timeout seconds.
-    timeout defaults to 1800s (30 min) for slow GPU node provisioning (A100 MIG).
+    Call begin_create_or_update, retrying IN_PROGRESS_CODES (a *previous* op still
+    in progress) up to MAX_RETRIES; poll until done, TimeoutError after
+    TIMEOUT_SECONDS.
 
-    Returns:
-        bool: True if a retry occurred, False if the operation succeeded on the first attempt.
+    Returns (request_started_at, retry_occurred): request_started_at is when the
+    accepted (2xx) attempt's PUT was issued, so callers count the accepted
+    request's frontend time and exclude the failed prior attempts as queue-wait.
+    On failure it attaches request_started_at to the exception (see below) so the
+    caller can still exclude the queue-wait.
     """
     retry_occurred = False
-    for attempt in range(retries):
+    for attempt in range(MAX_RETRIES):
+        request_started_at = time.time()
         try:
             poller = aks_sdk_client.agent_pools.begin_create_or_update(
                 resource_group_name=resource_group,
@@ -112,29 +134,38 @@ def begin_create_or_update_with_retry(
                 agent_pool_name=node_pool_name,
                 parameters=parameters,
             )
+        except HttpResponseError as e:
+            in_progress = any(code in str(e) for code in IN_PROGRESS_CODES)
+            if in_progress and attempt < MAX_RETRIES - 1:
+                retry_occurred = True
+                error_code = e.error.code if e.error else str(e)
+                logger.warning(
+                    f"Cluster has an in-progress operation, retrying in {RETRY_WAIT_SECONDS}s "
+                    f"(attempt {attempt + 1}/{MAX_RETRIES}): {error_code}"
+                )
+                time.sleep(RETRY_WAIT_SECONDS)
+                continue
+            if in_progress:
+                e.request_started_at = time.time()
+            raise
+        try:
             elapsed = 0
             while not poller.done():
-                time.sleep(poll_interval)
-                elapsed += poll_interval
-                if elapsed >= timeout:
+                time.sleep(POLL_INTERVAL_SECONDS)
+                elapsed += POLL_INTERVAL_SECONDS
+                if elapsed >= TIMEOUT_SECONDS:
                     raise TimeoutError(
-                        f"Node pool {node_pool_name} {label}timed out after {timeout}s"
+                        f"Node pool {node_pool_name} {label}timed out after {TIMEOUT_SECONDS}s"
                     )
                 logger.info(
                     f"Waiting for node pool {node_pool_name} {label}to complete "
                     f"({elapsed}s elapsed)..."
                 )
             poller.result()
-            return retry_occurred
-        except HttpResponseError as e:
-            if any(code in str(e) for code in ("OperationNotAllowed", "EtagMismatch")) and attempt < retries - 1:
-                retry_occurred = True
-                error_code = e.error.code if e.error else str(e)
-                logger.warning(
-                    f"Cluster has an in-progress operation, retrying in {retry_wait}s "
-                    f"(attempt {attempt + 1}/{retries}): {error_code}"
-                )
-                time.sleep(retry_wait)
-            else:
-                raise
-    return retry_occurred
+        except Exception as e:
+            e.request_started_at = request_started_at
+            raise
+        return request_started_at, retry_occurred
+    raise RuntimeError(
+        f"Node pool {node_pool_name} {label}exhausted {MAX_RETRIES} create/update retries"
+    )
--- a/modules/python/clients/aks_machine_client.py
+++ b/modules/python/clients/aks_machine_client.py
@@ -18,6 +18,7 @@ import requests
 from requests.adapters import HTTPAdapter
 
 from clients.aks_client import AKSClient
+from utils.common import compute_throttle_delay
 from utils.logger_config import get_logger, setup_logging
 
 # Configure logging.
@@ -505,6 +506,8 @@ class AKSMachineClient(AKSClient):
             machine_workers=machine_workers,
             timeout=timeout,
             batch_command_execution_times={},
+            chunk_throttle=[],
+            throttle_lock=threading.Lock(),
         )
         with self._get_operation_context()(
             "scale_machine", "azure", metadata, result_dir=self.result_dir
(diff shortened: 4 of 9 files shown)


modules/python/crud/operation.py
```diff
@@ -70,6 +86,17 @@ def end(self, success: bool = True, error: Optional[Exception] = None) -> None:
...
+                # operation to clear. That queue-wait is not part of this
+                # operation's latency, so shift the effective start forward
+                # (keeping start/end/duration consistent) and record the wait.
+                if self.excluded_seconds > 0:
+                    excluded = min(
+                        self.excluded_seconds, (end_dt - start_dt).total_seconds()
+                    )
+                    start_dt = start_dt + timedelta(seconds=excluded)
+                    self.start_timestamp = start_dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+                    self.metadata["in_progress_wait_seconds"] = self.excluded_seconds
```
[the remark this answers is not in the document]
xuexu6666: Good catch — fixed in 9e2a010e. `end()` now records the clamped `excluded` value in `in_progress_wait_seconds` so the metadata can never exceed the wall-clock elapsed and always matches the duration adjustment applied. Added a unit test (`test_operation_end_clamps_excluded_to_wall_clock`) covering the excess-wait edge case (duration floors at 0, metadata == wall-clock).

xuexu6666:
Ran a 3-reviewer pass over this PR and applied fixes (commit `64c400fc`):

**Fixed (issues multiple reviewers converged on):**
- **`in_progress_wait_seconds` double-write** — callers wrote the raw wait and `Operation.end()` then overwrote it with the clamped value (two writers, possible disagreement). `Operation.end()` is now the **single owner**; removed the caller-side writes in `instrument_nodepool_provisioning` and `scale_machine`, so the metadata always equals the amount actually removed from `duration`.
- **`union_seconds` hardening** — now drops inverted/zero-length intervals (`end <= start`) so a non-monotonic clock reading can never contribute negative time. Added tests (touching, inverted, zero-length).
- **`exclude_time` guard** clarified (`seconds is not None and seconds > 0`); `excluded_seconds` initialized as float.
- **Test coverage gaps** — added: budget-exhaustion re-raises the last error; `retries=0` → `RuntimeError` guard; asserted `command_execution_time` in the machine throttle test; `union_seconds` edge cases.
- **Documented** the batch-only scope of throttle exclusion (the individual PUT path uses opaque transport-level retries) and the concurrency ordering assumption in the mocked-`time` instrument tests.

**Deliberately not changed (rationale):**
- **Second-granularity timestamps** (`%Y-%m-%dT%H:%M:%SZ`): pre-existing across all Operations; changing precision has broad ADX/test blast radius, so out of scope for this PR (worth a separate follow-up).
- **`node_readiness_time` measured from `command_start`**: intentional — command and readiness share the same origin, so the ARM-vs-K8s `Delta` cancels it and stays correct while both exclude the queue-wait.
- **Cross-pipeline `command_execution_time` meaning** (AKS = submit+provision vs Machine = submit-only): pre-existing field-semantics divergence, not introduced here.

Tests green (pre-existing `test_log_gpu_mode_console_echo` is Python-3.10-only `assertNoLogs`), pylint clean, all modules under the 1000-line cap.

karenychen:
### Code review

Found 1 issue:

1. The union of every worker's 429 sleep is not necessarily time the concurrent scale command was blocked. If one batch worker sleeps while another is still actively submitting (and the active worker determines the overall completion time), subtracting that interval removes productive wall time from both `command_execution_time` and the entire operation duration. This can under-report Machine API latency; exclusion needs to be based on periods when the critical path/all unfinished workers are blocked, or measured per worker and composed by the actual completion critical path.

https://github.com/Azure/telescope/blob/8b6fbe6950fb405ffcade6669008bd54ebe48bca/modules/python/clients/aks_machine_client.py#L545-L555

xuexu6666:
@karenychen great catch — you're right, `union_seconds` under-reports latency: a 429 sleep on a worker that isn't on the completion critical path is hidden behind another worker's productive work and shouldn't be excluded at all. Fixed in `20603f86`.

Replaced the union with a **critical-path (makespan) attribution**. Each chunk's backoff is pure additive delay on its own timeline, so its throttle-free finish is `finish − backoff`. The command finishes at the makespan `max(finish)`; without throttle it would finish at `max(finish − backoff)`. The delay actually charged to the command is:

```
throttle_wait = makespan − max_i(finish_i − backoff_i)
```

This is exactly "measured per worker and composed by the actual completion critical path": off-critical-path backoff doesn't move the max, so it contributes 0; and when the critical chunk's backoff would drop it below the runner-up, the exclusion is correctly capped at the runner-up's finish (not the full backoff).

Changes:
- `_make_batch_request` now **returns its total 429 backoff** (0 if no retry) instead of recording raw sleep intervals.
- `_create_batch_machines` records `(finish_time, backoff)` per chunk.
- `scale_machine` computes the exclusion via `utils.throttle_makespan_delay(...)` (replaces `union_seconds`).

Tests cover the three cases: lone-chunk backoff counts fully; off-critical-path backoff → 0 (your scenario); critical-path backoff capped by the runner-up's throttle-free finish. (Individual PUTs are single-shot / no 429 retry, so their `chunk_throttle` is empty → 0.)

karenychen:
### Code review

Found 2 issues:

1. A batch chunk that exhausts its 429 retries never reaches the code that appends `(finish_time, backoff)` to `request.chunk_throttle`. `_scale_machine_batch` swallows that worker exception and returns a partial success list, so `scale_machine` computes the failed operation exclusion from successful chunks only. If the failed chunk spent 1+2+4 seconds backing off, those 7 seconds remain in `command_execution_time` and the operation duration—the exact queue wait this PR intends to remove. Preserve the accumulated backoff/finish data on the failure path as well.

https://github.com/Azure/telescope/blob/f814e6b5c3492163620102a99e2c76f28d605bba/modules/python/clients/aks_machine_client.py#L914-L925

2. Node-pool queue wait is also applied only after both concurrent futures succeed. When ARM retries one or more `OperationNotAllowed` responses and then the accepted operation fails, or when Kubernetes readiness fails after ARM succeeds, the early exception branch re-raises before reading `command_start` and calling `op.exclude_time`. Failed operation records therefore still include all pre-accept queue wait. Carry the accepted-attempt timing through the failure path when available and exclude it before re-raising.

https://github.com/Azure/telescope/blob/f814e6b5c3492163620102a99e2c76f28d605bba/modules/python/utils/provisioning_instrumentation.py#L53-L83

karenychen:
### Code review

Found 1 issue:

1. When every create/update attempt is rejected with `OperationNotAllowed` or `EtagMismatch`, the final-attempt branch re-raises the `HttpResponseError` without attaching any timing information. `instrument_nodepool_provisioning` therefore sees no `request_started_at` and excludes none of the preceding rejected-attempt round trips or backoff sleeps. With the default retry budget, a never-admitted operation still records roughly 270 seconds of queue wait in its failed-operation duration, contrary to the queue-wait exclusion. Preserve the timing anchor or accumulated wait on this terminal rejection path as well.

https://github.com/Azure/telescope/blob/16eedb7b860ba28bcfdfeedd21ff4898ce964aeb/modules/python/utils/provisioning_instrumentation.py#L151-L164

xuexu6666:
@karenychen fixed in `f319aa19` — thanks, you're right.

The terminal-rejection branch in `begin_create_or_update_with_retry` now anchors the exclusion at the give-up moment:

```python
# Terminal rejection (retries exhausted / never admitted, or non-retryable):
# the elapsed rejected attempts + backoff are all queue-wait.
e.request_started_at = time.time()
raise
```

`instrument_nodepool_provisioning` already reads `getattr(arm_exc, "request_started_at", None)` on the failure branch, so a never-admitted op now excludes the full rejected-attempt wait (~270s at the default budget) from its failed-operation duration instead of leaving it in. Anchoring at `now` also does the right thing for an immediate non-retryable failure — `now ≈ start`, so it excludes ~0 (nothing was waited).

Added `test_terminal_rejection_attaches_timing_anchor`. All three failure modes now exclude queue-wait consistently: accepted-then-failed, K8s-readiness-failed, and never-admitted.

karenychen:
### Code review

Found 1 issue:

1. This terminal branch also handles non-retryable `HttpResponseError`s (for example validation, auth, or not-found responses), but it always stamps `request_started_at` after `begin_create_or_update` returns. The caller consequently subtracts the entire failed request round trip and records it as `in_progress_wait_seconds`, even though no in-progress-operation retry or backoff occurred. A slow 400/403/404 can therefore collapse the failed operation duration toward zero and mislabel real ARM frontend latency as queue wait. Only use the give-up timestamp when the terminal error is the final `OperationNotAllowed`/`EtagMismatch`; non-retryable errors should retain the operation start/no exclusion.

https://github.com/Azure/telescope/blob/f319aa19348851151bf2672afc37b91a4568876d/modules/python/utils/provisioning_instrumentation.py#L151-L167

xuexu6666:
@karenychen fixed in `c2ca33c9` — good catch, the anchor was too broad.

The terminal branch now only anchors the exclusion when the error is an exhausted in-progress rejection; a non-retryable error attaches nothing:

```python
in_progress = any(code in str(e) for code in ("OperationNotAllowed", "EtagMismatch"))
if in_progress and attempt < retries - 1:
    ...  # retry
# Terminal:
if in_progress:                       # exhausted OperationNotAllowed/EtagMismatch
    e.request_started_at = time.time()  # all queue-wait -> exclude
raise                                 # non-retryable (400/403/404/...) -> no anchor, real ARM latency retained
```

So a slow 400/403/404 keeps its full round trip in the failed-op duration and is not mislabeled as `in_progress_wait_seconds`. Added an assertion to `test_raises_non_retryable_error` that no timing anchor is attached.

karenychen:
### Code review

Found 1 issue:

1. `throttle_makespan_delay` computes how much 429 backoff delayed only the concurrent PUT phase, but `op.exclude_time(throttle_wait)` removes that amount from the entire scale operation through agent-pool provisioning and node readiness. Those phases can have a different critical path: for example, chunk A may finish its PUT last solely because of a 5s backoff, while a machine from chunk B becomes the last Ready node. Removing chunk A backoff then shortens the recorded end-to-end provisioning duration even though it did not delay completion. Keep this exclusion scoped to `command_execution_time`, or derive end-to-end exclusion from the chunk that actually determines operation completion.

https://github.com/Azure/telescope/blob/c2ca33c999bd5e0e0d04633a054f0c7c7ace4eee/modules/python/clients/aks_machine_client.py#L545-L555

xuexu6666:
@karenychen fixed in `91557882` — agreed, the whole-duration exclusion was wrong for the concurrent case.

Scoped the throttle exclusion to `command_execution_time` only; the operation duration is no longer adjusted:

```python
throttle_wait = throttle_makespan_delay(request.chunk_throttle)
op.add_metadata("command_execution_time", max(0.0, command_end - command_t0 - throttle_wait))
op.add_metadata("throttle_wait_seconds", throttle_wait)   # recorded for visibility
# (op.exclude_time is no longer called on the machine path)
```

Rationale: `throttle_makespan_delay` measures the throttle's delay to the **PUT-phase** makespan, which is correct for `command_execution_time`. But as you noted, the PUT-phase critical path can differ from the end-to-end one — the last-Ready machine may be a different chunk — so PUT-phase throttle doesn't necessarily delay completion, and removing it from the whole operation duration can shorten it incorrectly. Deriving the true end-to-end exclusion would require correlating each chunk's backoff to the specific machine that determines readiness completion, which isn't reliably available here, so the conservative scope is the right call. The throttle is still visible as `throttle_wait_seconds`.

Updated the test to assert `command_execution_time`/`throttle_wait_seconds` are set and `op.exclude_time` is NOT called on the machine path.

Note this makes the two pipelines intentionally asymmetric: the AKS node-pool path still excludes queue-wait from the duration (single PUT -> the pre-accept wait strictly delays everything downstream, one critical path), whereas the concurrent Machine path scopes it to command time only.

karenychen:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

djsly:
exclude the failed-attempt wait (shared Operation.exclude_time primitive
 
 is this getting compute somewhere else now ?

xuexu6666:
@djsly good catch — that line in the description was stale (it drifted across the review rounds). Current state at `91557882`:

- **AKS node-pool path**: still uses the shared `Operation.exclude_time()` primitive — the failed-attempt / queue-wait is excluded from the operation **`duration`** and recorded as `in_progress_wait_seconds`. (Also on the failure paths.)
- **Machine path**: after @karenychen's feedback, the throttle exclusion was **removed from `Operation.exclude_time`** — subtracting PUT-phase throttle from the *whole-operation* duration was wrong when the end-to-end critical path differs (the last-Ready machine may be a different chunk than the one whose PUT was throttled). It's now computed **inline into `command_execution_time` only**:

```python
throttle_wait = throttle_makespan_delay(request.chunk_throttle)   # makespan − max(finish − backoff)
op.add_metadata("command_execution_time", max(0.0, command_end - command_t0 - throttle_wait))
op.add_metadata("throttle_wait_seconds", throttle_wait)
# no op.exclude_time on the machine path -> operation duration is unchanged
```

`grep exclude_time modules/python/clients/aks_machine_client.py` → 0 hits, confirming it's not computed elsewhere for the machine path; the duration simply isn't throttle-adjusted there anymore.

I've refreshed the PR description to reflect this (the old text still mentioned `union_seconds` and duration-exclusion for the machine path — both superseded).

modules/python/utils/common.py
```diff
@@ -71,3 +72,30 @@ def str2bool(val):
     if val.lower() in ("false", "no", "0"):
         return False
     raise argparse.ArgumentTypeError("Boolean value expected.")
+
+
+def throttle_makespan_delay(chunks: List[Tuple[float, float]]) -> float:
```
liyu-ma: The name of the function is very confusing. I feels it is better named `compute_throttle_delay`.

modules/python/utils/common.py
```diff
@@ -71,3 +72,30 @@ def str2bool(val):
...
+
+    Precondition: all workers must have STARTED concurrently (no slot queueing), so
+    that ``finish - backoff`` is a worker's true throttle-free finish. The Machine
+    caller satisfies this -- it submits exactly ``machine_workers`` chunks into a
+    ``max_workers=machine_workers`` pool, so every chunk gets its own slot. If chunks
+    were ever queued behind slots this attribution would under-count the delay.
+    """
+    if not chunks:
+        return 0.0
+    makespan = max(finish for finish, _ in chunks)
```
liyu-ma: suggest renaming:
makespan -> `actual_finish`
counterfactual -> `finish_without_backoff`

modules/python/utils/provisioning_instrumentation.py
```diff
@@ -61,23 +61,45 @@ def instrument_nodepool_provisioning(
...
+        # Even on failure, exclude the pre-accept queue-wait from the recorded
+        # duration when we know when the accepted attempt started: from the ARM
+        # result if ARM succeeded (K8s failed), else from the timestamp the retry
+        # helper attached to the exception (accepted, then the op failed/timed out).
+        command_start = None
+        if not arm_exc:
+            (command_start, _), _ = arm_future.result()
+        else:
+            command_start = getattr(arm_exc, "request_started_at", None)
+        if command_start is not None:
```
liyu-ma: I see there are two places in this function to calc the excluded time, can you make small helper like
```
def _record_excluded_time(op, command_start, start_time):
    if command_start is not None:
        op.exclude_time(max(0, command_start - start_time))
```

and use it for line 73 and 87

modules/python/utils/provisioning_instrumentation.py
```diff
@@ -61,23 +61,45 @@ def instrument_nodepool_provisioning(
...
+
+    # exclude_time feeds the operation duration; Operation.end() records the
+    # (clamped) excluded amount as in_progress_wait_seconds -- single owner of
+    # that metadata key, so it always matches the duration adjustment applied.
+    op.exclude_time(in_progress_wait)
     op.add_metadata("node_readiness_time", node_readiness_time)
     op.add_metadata("command_execution_time", command_execution_time)
-    op.add_metadata("retry_occurred", bool(arm_result))
+    op.add_metadata("retry_occurred", retry_occurred)
     logger.info(
```
liyu-ma: Now this logger line gets even inflated. Can we make it a small helper:
```
def _log_provisioning_success( ... )
```

modules/python/utils/provisioning_instrumentation.py
```diff
@@ -100,18 +122,55 @@ def begin_create_or_update_with_retry(
...
+        request_started_at = time.time()
         try:
             poller = aks_sdk_client.agent_pools.begin_create_or_update(
                 resource_group_name=resource_group,
                 resource_name=cluster_name,
                 agent_pool_name=node_pool_name,
                 parameters=parameters,
             )
+        except HttpResponseError as e:
+            in_progress = any(code in str(e) for code in ("OperationNotAllowed", "EtagMismatch"))
```
liyu-ma: can we put these 2 error types to a module constant like
`IN_PROGRESS_CODES `

modules/python/utils/provisioning_instrumentation.py
liyu-ma: From line 115-118:
I don't see any variant values real usage of default params. Since you are touching this method, can I request you to remove these 4 and use constants instead:
```
MAX_RETRIES = 10                 # in-progress (409) submit attempts
RETRY_WAIT_SECONDS = 30      # sleep between rejected submit attempts
POLL_INTERVAL_SECONDS = 30   # poller.done() polling cadence
TIMEOUT_SECONDS = 1800
```
Tests may need to be updated.

liyu-ma, line 1: A general comment:
I see more verbose comments are added in. Personally I am against it - prefer readability via functions / variables naming. Can you try either reduce them or make them as succinct as possible

modules/python/utils/common.py
xuexu6666: Done in `61ba4ec6` — renamed to `compute_throttle_delay`.

modules/python/utils/common.py
xuexu6666: Done — `makespan` -> `actual_finish`, `counterfactual` -> `finish_without_backoff`.

modules/python/utils/provisioning_instrumentation.py
xuexu6666: Done — extracted `_record_excluded_time(op, command_start, start_time)` (returns the excluded seconds) and used it on both the failure branch and the success path.

modules/python/utils/provisioning_instrumentation.py
xuexu6666: Done — extracted `_log_provisioning_success(...)`; the success path just calls it now.

xuexu6666:
Addressed the readability feedback in `61ba4ec6` (all of @liyu-ma's threads resolved):

- Renamed `throttle_makespan_delay` → `compute_throttle_delay`; locals `makespan`/`counterfactual` → `actual_finish`/`finish_without_backoff`.
- Extracted `_record_excluded_time(op, command_start, start_time)` (used on both the success path and the failure branch) and `_log_provisioning_success(...)`.
- Trimmed the verbose comments across `operation.py` / `provisioning_instrumentation.py` / `aks_machine_client.py` / `common.py` — leaning on the names/helpers instead of prose. Net −~30 comment lines (`aks_machine_client.py` 998 → 979).

Tests green (only the pre-existing Py3.10-only `assertNoLogs`), pylint 10/10.

modules/python/utils/provisioning_instrumentation.py
xuexu6666: Done in `568f8f39` — added module constant `IN_PROGRESS_CODES = ("OperationNotAllowed", "EtagMismatch")` and use it in the retry check.

modules/python/utils/provisioning_instrumentation.py
xuexu6666: Done — removed the 4 defaulted params; the function now uses module constants `MAX_RETRIES`/`RETRY_WAIT_SECONDS`/`POLL_INTERVAL_SECONDS`/`TIMEOUT_SECONDS`. The single production caller passed none of them. Updated the tests to patch those constants where a non-default is needed (e.g. `@mock.patch("...MAX_RETRIES", 3)`).

xuexu6666: Addressed in the prior commit `61ba4ec6` — trimmed the verbose comments across operation.py / provisioning_instrumentation.py / aks_machine_client.py / common.py and extracted helpers (`_record_excluded_time`, `_log_provisioning_success`) + better names, leaning on those instead of prose.

liyu-ma (approved):
LGTM

PR #1307, opened 2026-09-02
Make kperf ingest batch names predictable and avoid residue on retry

xinWeiWei24:
Drop the -first/-retry name suffix so each batch uses a stable set name "<name>-b<batch>", making object names predictable for kperf load profiles to target by prefix. Replace the rename-based retry (which left partial objects from the failed attempt as orphans) with a bounded retry loop that deletes the partial batch before re-creating, so each batch ends up with exactly COUNT objects and accurate progress accounting.


kcl/lib/steps/kperf/data_ingest.k
```diff
@@ -38,14 +41,24 @@ for batch in $(seq 1 $TOTAL_BATCHES); do
...
+
+  # add is not idempotent, so clear the partial batch before each retry.
+  for attempt in $(seq 1 ${MAX_ATTEMPTS}); do
+    run_batch && break
+    if [ "$attempt" -eq "${MAX_ATTEMPTS}" ]; then
+      echo "Batch $batch failed after ${MAX_ATTEMPTS} attempts" >&2
+      exit 1
+    fi
+    echo "Batch $batch attempt $attempt failed; cleaning up before retry..."
+    runkperf data ${dataType} --namespace ${namespace} delete ${name}-b$batch || true
```
liyu-ma: I think you need to pass in the `total` param here, otherwise the default target count is 10
https://github.com/Azure/kperf/blob/main/contrib/cmd/runkperf/commands/data/configmaps/configmap.go#L154-L158

liyu-ma (approved):

PR #1309, opened 2026-09-02
Fix hyperscale pod scheduling capacity

liyu-ma:
## Summary
- size the physical KWOK node pool from the virtual-node count and nodes-per-controller ratio
- generate 20, 40, and 80 KWOK pool nodes for the H2, H4, and H8 stages respectively
- flush KWOK controller readiness messages immediately so Azure DevOps timestamps reflect actual polling time
- regenerate the pod-scheduling pipeline YAML

## Why
The H4 benchmark created 40 KWOK controller Deployments but retained a fixed 20-node `Standard_D4_v3` pool. Only 20 controller Pods could schedule, and controller readiness timed out. H8 would have the same issue with 80 controllers.

## Validation
- regenerated `kcl/ccp_team/hyperscale_pod_scheduling/pipeline.yaml`
- verified generated KWOK pool counts are H2=20, H4=40, H8=80
- ran focused KWOK Python tests successfully
- ran Python compilation and `git diff --check` successfully

Replaces closed PR #1308, which targeted the wrong base branch.


xinWeiWei24 (approved):
lgtm

wonderyl (approved):

PR #1312, opened 2026-09-04
feat(kperf): support exact count and configurable batch size

xinWeiWei24:
Allow callers to specify the exact object count and batch size while preserving the existing GiB-based ingestion behavior.

### DIFF
--- a/kcl/lib/steps/kperf/data_ingest.k
+++ b/kcl/lib/steps/kperf/data_ingest.k
@@ -57,12 +57,8 @@ for batch in $(seq 1 $TOTAL_BATCHES); do
       exit 1
     fi
     echo "Batch $batch attempt $attempt failed; cleaning up before retry..."
-<<<<<<< HEAD
     # delete reconstructs names <set>-<0..total-1>, so total must cover the whole batch.
     runkperf data ${dataType} --namespace ${namespace} delete ${name}-b$batch --total $COUNT || true
-=======
-    runkperf data ${dataType} --namespace ${namespace} delete ${name}-b$batch || true
->>>>>>> 70e38325 (Make kperf ingest batch names predictable and avoid residue on retry)
   done
 
   completed=$((completed + COUNT))
--- a/kcl/lib/steps/kperf/data_ingest.k
+++ b/kcl/lib/steps/kperf/data_ingest.k
@@ -5,34 +5,40 @@ DEFAULT_QPS = 80
 DEFAULT_BURST = 80
 DEFAULT_SIZE_BYTES = 524288  # 0.5 MiB per object
 DEFAULT_GROUP_SIZE = 80
-BATCH_SIZE = 2000  # Internal batch size, not configurable by caller
 MAX_ATTEMPTS = 2  # Attempts per batch before failing the step
 BYTES_PER_GIB = 1073741824
 
-# RunKperfDataIngest creates objects in fixed-size batches (BATCH_SIZE) to be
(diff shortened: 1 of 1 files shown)


kcl/lib/steps/kperf/data_ingest.k
```diff
@@ -5,34 +5,40 @@ DEFAULT_QPS = 80
...
+#
+# Batching is used to be
 # resilient to failures — if a batch fails, only that batch needs to be retried, and the
 # local in-memory counter stays accurate since each successful batch represents a known
 # quantity. This avoids using kubectl get to count existing objects, which would OOM
 # the pipeline agent by fetching full object bodies.
 # Each batch uses the set name "<name>-b<batch>", so object names are predictable and
 # kperf load profiles can target a batch by its prefix.
-RunKperfDataIngest = lambda dataType: str, namespace: str, name: str, totalGiB: str, qps: int = DEFAULT_QPS, burst: int = DEFAULT_BURST, size: int = DEFAULT_SIZE_BYTES, groupSize: int = DEFAULT_GROUP_SIZE -> steps.Step {
+RunKperfDataIngest = lambda dataType: str, namespace: str, name: str, totalGiB: str = "0", qps: int = DEFAULT_QPS, burst: int = DEFAULT_BURST, size: int = DEFAULT_SIZE_BYTES, groupSize: int = DEFAULT_GROUP_SIZE, total: int = 0, batchSize: int = 2000 -> steps.Step {
```
wonderyl: Should disable totalGiB and total are both set?

wonderyl: move total and totalGiB closer, so that reader can easily spot them together.

xinWeiWei24: Done. total and totalGiB are now next to each other in the parameter list.

xinWeiWei24: This is already handled: when both are set, total takes precedence and totalGiB is ignored.

wonderyl: Yon only commented, not enforced. Use assert and raise and error if both are set.

xinWeiWei24: Thanks. Previously, lines 30-32 made total take precedence when both were set. I’ve now added an assertion so setting both is rejected instead.

kcl/lib/steps/kperf/data_ingest.k
```diff
@@ -20,6 +20,7 @@ BYTES_PER_GIB = 1073741824
 # Each batch uses the set name "<name>-b<batch>", so object names are predictable and
 # kperf load profiles can target a batch by its prefix.
 RunKperfDataIngest = lambda dataType: str, namespace: str, name: str, totalGiB: str = "0", total: int = 0, qps: int = DEFAULT_QPS, burst: int = DEFAULT_BURST, size: int = DEFAULT_SIZE_BYTES, groupSize: int = DEFAULT_GROUP_SIZE, batchSize: int = 2000 -> steps.Step {
+  assert total == 0 or totalGiB == "0", "total and totalGiB cannot both be set"
```
wonderyl, line 23: Do you need to enforce one of them is set as well? given the compiler won't enforce it.

xinWeiWei24: Line 79 already skips the step when both are unset. This makes it easier to reuse in repeated scenario steps where ingestion is optional. An assert would fail KCL generation instead of skipping the step.

wonderyl (approved):

PR #1318, opened 2026-09-09
Fix azure SDK namespace shadowing in CRUD pipeline steps (run via -m crud.main)

xuexu6666:
## Problem

The **GPU Cluster CRUD** pipeline (and other CRUD/Machine-API pipelines) have failed on every run since #1231 merged. Both CRUD tasks fail immediately:

```
File ".../modules/python/crud/main.py", line 19
  from crud.azure.node_pool_crud import NodePoolCRUD
File ".../modules/python/clients/aks_client.py", line 20
  from azure.core.exceptions import HttpResponseError, ResourceNotFoundError
ModuleNotFoundError: No module named 'azure.core'
```

`azure-core` **is** installed correctly by "Install Dependencies" — this is namespace shadowing, not a missing dependency.

## Root cause

The steps run the script as a **file**:

```yaml
workingDirectory: modules/python
env:
  PYTHON_SCRIPT_FILE: $(Pipeline.Workspace)/s/modules/python/crud/main.py
script: python3 "$PYTHON_SCRIPT_FILE" create ...
```

Running `python3 .../crud/main.py` puts the script's own directory — `modules/python/crud/` — at `sys.path[0]`. #1231 added `modules/python/crud/azure/__init__.py`, turning `crud/azure/` into a **regular package**. Found first on `sys.path`, it shadows the installed `azure` SDK namespace package, so `azure.core` can no longer be resolved.

Unit tests didn't catch this because pytest runs with a different `sys.path` that never places `crud/` at `sys.path[0]`.

## Fix

Invoke the module with `-m` instead of running the file. `python3 -m crud.main` puts `modules/python` (the working directory) on `sys.path[0]` instead of `crud/`, so the `azure` SDK resolves from site-packages while `crud.azure` still imports. This is also what the `telescope-crud` entry point registered in #1231 was intended for. The now-unused `PYTHON_SCRIPT_FILE` env var is removed.

Applied to all affected engine steps:
- `steps/engine/crud/k8s/execute.yml`
- `steps/engine/crud/k8s/collect.yml`
- `steps/engine/machine/k8s/execute.yml`
- `steps/engine/machine/k8s/collect.yml`

## Verification

Reproduced the shadowing and confirmed the fix locally:

```
# crud/ on sys.path[0] (file invocation)
import azure.core.exceptions  -> ModuleNotFoundError: No module named 'azure.core'

# modules/python on sys.path[0] (-m crud.main)
import azure.core.exceptions; import crud.azure  -> both import OK
```

### DIFF
--- a/steps/engine/crud/k8s/collect.yml
+++ b/steps/engine/crud/k8s/collect.yml
@@ -9,12 +9,11 @@ steps:
 - script: |
     set -eo pipefail
 
-    PYTHONPATH=$PYTHONPATH:$(pwd) python3 $PYTHON_SCRIPT_FILE collect
+    PYTHONPATH=$PYTHONPATH:$(pwd) python3 -m crud.main collect
 
   displayName: 'Collect K8s CRUD Operations data for ${{ parameters.cloud }}'
   workingDirectory: modules/python
   env:
-    PYTHON_SCRIPT_FILE: $(Pipeline.Workspace)/s/modules/python/crud/main.py
     RESULT_DIR: $(System.DefaultWorkingDirectory)/$(RUN_ID)
     RUN_URL: $(System.TeamFoundationCollectionUri)$(System.TeamProject)/_build/results?buildId=$(Build.BuildId)&view=logs&j=$(System.JobId)
     REGION: ${{ parameters.region }}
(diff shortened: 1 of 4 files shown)


xuexu6666:
Superseded by #1319 (branch now on Azure/telescope directly, no fork).

PR #1320, opened 2026-09-10
Configure KWOK controller resource requests and limits

liyu-ma:
## Why

KWOK deployments may need to size down to one virtual node per controller. The existing hardcoded controller requests and limits are excessive at this 1:1 density, so benchmark pipelines need a way to right-size each controller.

## Changes

- Add CLI options for controller CPU and memory requests and limits.
- Preserve the existing resource values as defaults for backward compatibility.
- Apply configured values to each generated KWOK controller Deployment.
- Add coverage for custom Kubernetes resource quantities.

## Validation

- All 12 KWOK Python tests pass.
- Python compilation and workspace diagnostics pass.


xinWeiWei24 (approved):
LGTM

wonderyl (approved):

PR #1324, opened 2026-09-14
(feat.) Add reusable Kubernetes Job pipeline step

liyu-ma:
## Summary

- Add a helper function `WaitForK8sJobFinished` under `kcl/lib/steps/k8s`, which can be used to wait for a k8s job to reach terminal state.
- The function polls `Complete` and `Failed` conditions with a bounded timeout and defaulted status counters
- It also emits container logs and Kubernetes diagnostics on failure

## Validation

- `kcl fmt kcl/lib/steps/k8s/wait_for_k8s_job.k`
- `kcl run kcl/lib/steps/k8s/wait_for_k8s_job.k`
- extracted embedded script validated with `bash -n`
- `git diff --check`


kcl/lib/steps/k8s/run_k8s_job.k
```diff
@@ -0,0 +1,49 @@
...
+    steps.Bash {
+        displayName = "Run Kubernetes Job: ${jobName}"
+        bash = """
+set -euo pipefail
+
+NAMESPACE="${namespace}"
+JOB_NAME="${jobName}"
+MANIFEST_YAML="${manifestYaml}"
+
+namespace_resource=$(kubectl get namespace "$NAMESPACE" --ignore-not-found -o name)
```
wonderyl: see comment https://msazure.visualstudio.com/DefaultCollection/CloudNativeCompute/_git/telescope/pullRequest/17154944?discussionId=261335792#1789458657

kcl/lib/steps/k8s/run_k8s_job.k
```diff
@@ -0,0 +1,49 @@
...
+
+kubectl apply -f "$MANIFEST_YAML"
+if ! kubectl -n "$NAMESPACE" get job "$JOB_NAME" >/dev/null; then
+    echo "##[error]Manifest did not create expected Job $NAMESPACE/$JOB_NAME"
+    exit 1
+fi
+
+deadline=$(( $(date +%s) + ${timeoutSeconds} ))
+final=""
+while [ $(date +%s) -lt $deadline ]; do
```
xinWeiWei24: not sure if we can leverage https://github.com/Azure/telescope/blob/v2/kcl/lib/steps/health/validate_job.k#L54 here

liyu-ma: The key difference is that I want to wait for the job to end, but `validate_job.k` only does the check once. It feels that they are for different use cases. The common part the shell script for checking the condition of pod/job, but I don't feel that is easy (or worth) to share among the kcl steps.

xinWeiWei24 (approved):

wonderyl (approved):

PR #1326, opened 2026-09-16
support optional H2/H4/H8 control plane scaling size

xinWeiWei24:
Add a scaling_size pipeline parameter (H2/H4/H8, default H8) so the control plane scaling size is selectable at run time instead of being hardcoded to H8. Add matching cl2-h2.yaml/cl2-h4.yaml manifests (cl2.yaml renamed to cl2-h8.yaml) so each size runs its own clusterloader2 config.

Also switch cluster creation from az rest to az aks create.


vittoriasalim (approved):

kcl/ccp_team/hyperscale_apiserver_executing_seats/cl2-h4.yaml
```diff
@@ -0,0 +1,555 @@
...
+    SMALL_STATEFUL_SETS_PER_NAMESPACE: 0
+    MEDIUM_STATEFUL_SETS_PER_NAMESPACE: 0
+    CL2_SMALL_JOBS_PER_NAMESPACE: 0
+    CL2_MEDIUM_JOBS_PER_NAMESPACE: 0
+    CL2_BIG_JOBS_PER_NAMESPACE: 0
+    CL2_RATE_LIMIT_POD_CREATION: false
+    CL2_EXIT_AFTER_EXEC: true
+    CL2_ENABLE_PVS: false
+    CL2_ENABLE_CLUSTER_OOMS_TRACKER: false
+    CL2_DEFAULT_QPS: 3600
```
wonderyl: Is it intentional that CL2_DEFAULT_QPS is the same for h2 and h4, but h8 is diffferent?

kcl/ccp_team/hyperscale_apiserver_executing_seats/cl2-h8.yaml
```diff
@@ -27,8 +28,8 @@ metadata:
   name: cl2
   namespace: clusterloader2
 spec:
-  completions: 8
-  parallelism: 8
+  completions: 6
```
wonderyl, line 36: the completion of h2 is 2, h4 is 4, while h8 is 6, what's the reasoning about it?

xinWeiWei24: 6 is chosen mainly to generate enough load without runs taking too long — in practice it's already sufficient and doesn't cause CCP or cl2 errors. It keeps each tier's control plane driven near saturation without the load drivers themselves becoming the bottleneck. Setting it to 8 would be a bit too much; the other tiers might also be slightly over, but not by much.

kcl/ccp_team/hyperscale_apiserver_executing_seats/cl2-h4.yaml
xinWeiWei24: Good catch — that was unintentional. I've updated all tiers to use CL2_DEFAULT_QPS: 4000 so H2/H4/H8 are consistent now.

kcl/ccp_team/hyperscale_apiserver_executing_seats/pipeline.k
```diff
@@ -4,17 +4,16 @@ import lib.const
...
-CONTROL_PLANE_SCALING_SIZE = "H8"
+LOCATION = "swedencentral"
+CONTROL_PLANE_SCALING_SIZE = "\${{ parameters.scaling_size }}"
 CLUSTER = "stg-${CONTROL_PLANE_SCALING_SIZE}-hyperscale-apiserver-seats"
 
 CL2_POOL = azure.NodePool {
     name = "cl2pool"
-    sku = "Standard_D16S_v4"
+    sku = "Standard_D16dS_v4"
     count = 10
```
liyu-ma: I don't think you need 10 nodes, if `completions` is max at 6? Can this be dynamic adjusted based on tier

kcl/ccp_team/hyperscale_apiserver_executing_seats/cl2-h8.yaml
wonderyl: could you put all the reasoning into comments? For anything that's not explicitly expressed in the code, you need to add comments.

kcl/ccp_team/hyperscale_apiserver_executing_seats/pipeline.k
```diff
@@ -26,63 +25,52 @@ KWOK_POOL = azure.NodePool {
...
+    --node-vm-size Standard_D8dS_v4 \\
+    --max-pods 250 \\
+    --network-plugin azure \\
+    --network-plugin-mode overlay \\
+    --pod-cidr 10.64.0.0/10 \\
+    --service-cidr 10.0.0.0/16 \\
+    --dns-service-ip 10.0.0.10 \\
+    --outbound-type managedNATGateway \\
+    --nat-gateway-managed-outbound-ip-count 10 \\
+    --tags SkipAKSCluster=true SkipASMAzSecPackAutoConfig=true SkipLinuxAzSecPack=true \\
```
liyu-ma: oh btw I think we are required to remove `SkipASMAzSecPackAutoConfig` and `SkipLinuxAzSecPack` tags?

wonderyl (approved):

liyu-ma (approved):

PR #1327, opened 2026-09-16
Regenerate pipeline YAMLs from KCL sources

wonderyl:
## Summary
- Regenerate pipeline YAMLs from KCL sources for all pipelines that build

## Test plan
- [x] `kcl run` succeeds for each regenerated pipeline


xinWeiWei24 (approved):

LeonardCareer (approved):
lgtm

PR #1328, opened 2026-09-16
Always run ASI dashboard step and regenerate pipeline YAMLs

wonderyl:
## Summary
- Add `condition="always()"` to the ASI dashboard step so its URL is printed even when prior steps fail
- Regenerate pipeline YAMLs from KCL sources

## Test plan
- [x] `kcl run` succeeds for all pipelines


xinWeiWei24 (approved):
LGTM

LeonardCareer (approved):
lgtm

